PR #175, opened 2021-05-11
Hotfix: persist change fee address in case of Taxi transaction

### LINKED ISSUE #173: Unknown blind key error when sending LBTC and paying with Taxi
image

louisinger:
The address using to send taxi's payout change is now persisted in wallet's store. It fixes "unknow blinding key" thrown at confirm step.

it closes #173 

@tiero please review


src/presentation/wallet/send/choose-fee.tsx
```diff
@@ -130,11 +131,8 @@ const ChooseFee: React.FC = () => {
           try {
             const wallet: IWallet = { ...wallets[0] };
             const feeChangeAddress = await nextAddressForWallet(wallet, app.network.value, true);
-            dispatch(
-              setFeeChangeAddress(
-                Address.create(feeChangeAddress.value, feeChangeAddress.derivationPath)
```
tiero, line 135: Why we do no use Address.create anymore'

Issue #116, opened 2021-03-15
[Login] Restore account should open fullscreen page


tiero:
At the moment it shows the restor flow inside the shell popup﻿

louisinger:
Do u mean "onboarding restorer flow"? It looks like this issue is closed if yes

Janaka-Steph:
Yes, the Restore button on popup login screen had to open full screen onboarding restorer flow.

tiero:
no, we need to activate a link/button in the bottom as per design, in the popup Login page, to force restore an account with seed


![image](https://user-images.githubusercontent.com/3596602/134318795-3905e00d-ac96-4558-a1d9-faefe860f5dd.png)


At the moment only way is to delete the extension and re-install

PR #178, opened 2021-05-19
Refactor with webext-redux and redux-persist 

### LINKED ISSUE #161: Transactions: Recognize swap transactions
Use this log icon for the transaction list and the label "Atomic swap" https://app.zeplin.io/project/5fd1e92cc5fa144b9c8f7c71/screen/609a677ab0bb433d1358e7ab

### LINKED ISSUE #167: Refactor with Redux proxy store

### LINKED ISSUE #177: create password page, miss an "f"
On the create password page, it's written "terms oS service" instead of "terms oF service"

### LINKED ISSUE #179: Tap on Send from Home when one of the assets is >999 crashes cause bad handling of localeString en-US
```
presentation.eac17017.js:16 TypeError: Cannot read property 'toLowerCase' of undefined
    at presentation.eac17017.js:2483
    at Array.filter (<anonymous>)
    at presentation.eac17017.js:2483
    at as (presentation.eac17017.js:16)
    at exports.unstable_runWithPriority (presentation.eac17017.js:12)
    at fa (presentation.eac17017.js:16)
    at ns (presentation.eac17017.js:16)
    at presentation.eac17017.js:16
    at R (presentation.eac17017.js:12)
```﻿

### LINKED ISSUE #181: Home: Asset list should have Bitcoin always on top

### LINKED ISSUE #183: Investigate "Invalid extended public key!" error
image

### LINKED ISSUE #185: [Bug] If extension is disabled and re-enabled again, it will not open anymore
Screenshot 2021-06-08 at 16 10 46

### PR DESCRIPTION
**Pull request by louisinger**
### This PR refactors Marina using [webext-redux](https://github.com/tshaddix/webext-redux) and [redux-persit](https://github.com/rt2zz/redux-persist)

**Major changes**

- drop the former "AppContext's redux" --> replace by web-ext redux.
- drop the repository layer --> replace by redux-persist.
- refactor the domain layer: as the class paradigm isn't adapted to redux, I rewrite the domains as TS interfaces.
- drop all the Thunk part --> all the functionnal logic is now done outside of redux's store.
- the "connected" data are now a reducer in redux store.
- hot reloading :champagne: :champagne: 
- update the transactions view: it closes #161 

**New architecture**

* The "real" redux store runs in background-script. it is wrapped to proxy stores (`web-ext` stores instanciated in UI components) in order to link the background and the proxy states. It allows background and UI contexts to access the same shared state (and mutate it via actions).

* We are using [web-ext aliases](https://github.com/tshaddix/webext-redux#4-optional-implement-actions-whose-logic-only-happens-in-the-background-script-we-call-them-aliases) to handle async updating. They redirect UPDATE_UTXOS, UPDATE_TXS and UPDATE_ASSETS actions to `Backend` functions dispatching asynchronously the update actions.

* The backgorund store is persisted in browser local storage (serialization format has not changed).

backup(2)

**Tests**

I have deleted a lot of unit tests (mainly related to former `AppContext` architecture). So the coverage is reduced but the amount of code to test has decreased. 

_functionnal testing_

- [x] send a withdraw transaction.
- [x] receive fund via deposit transaction.
- [x] send a withdraw transaction with Taxi.
- [x] send a self-transfer transaction.
- [x] check if utxos is updated after receive/send.
- [x] check if transactions list is updated after receive/send.
- [x] swap transaction is OK.
- [x] build and install the extension on chrome and firefox 
- [x] window.marina API 
- ???

--
please review @tiero @Janaka-Steph 

it closes #167 
it closes #177 
it closes #181 
it closes #183 
it closes #179 
it closes #185

### CODE REVIEW

**Comment by tiero** [comment]
can you solve conflicts? @louisinger

**Comment by Janaka-Steph** [comment]
Sending L-BTC paying fees in USDT fails with `unknow blinding key for script 0014103ca3145f879ae243f3d4034055c100f16957e9`

**Comment by Janaka-Steph** [comment]
First time we click on USDt button in Choose Fee page, it doesn't take effect; we see a glitch, the app re-render. This is because the `balances` prop is recomputed. Even though the object properties remain the same, the object is different (referential equality). Then it triggers the useEffect line 130 `setFeeCurrency(supportedAssets[0])`, setting the feeCurrency to L-BTC again instead of USDT. 
To avoid triggering the useEffect, simply set the balances prop in a state variable that will keep referential equality.
const [balancesVar] = useState(balances);

**Review by tiero — CHANGES_REQUESTED** [review]

File: src/application/background-script.ts [hunk 1]
```diff
@@ -22,35 +30,14 @@ browser.runtime.onInstalled.addListener(({ reason }) => {
     switch (reason) {
       //On first install, open new tab for onboarding
       case 'install':
-        await initPersistentStore(repos);
-
-        // this is for development only
-        if (process.env.SKIP_ONBOARDING) {
```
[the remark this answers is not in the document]
louisinger: with hot-reloading, we don't need to onboard each time we save a file so IMO no need to add something to code only to save the first onboarding time but as u want, I don't have a real preference here

src/application/background-script.ts
```diff
@@ -22,35 +30,14 @@ browser.runtime.onInstalled.addListener(({ reason }) => {
...
-      //   {
-      //     const url = browser.runtime.getURL('updated.html');
-      //     browser.tabs.create({ url }).catch(console.log);
-      //   }
-      //   break;
     }
   })().catch(console.error);
 });
 
-browser.runtime.onStartup.addListener(() => {
```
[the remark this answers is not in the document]
louisinger: duplicate the call L60

src/application/background-script.ts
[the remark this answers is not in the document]
tiero: Yes was there for dev. Maybe leave it for now, we will drop later on maybe

package.json
```diff
@@ -35,7 +35,7 @@
     "watchskip": "npm run clean && npm run copy-assets && SKIP_ONBOARDING=true run-p watch:**",
     "watch": "npm run clean && npm run copy-assets && run-p watch:**",
     "watch:html": "NETWORK=regtest parcel watch public/*.html",
-    "watch:scripts": "NETWORK=regtest parcel watch src/application/inject.ts src/application/background-script.ts src/application/content-script.ts",
```
tiero, line 38: no use of NETWORK at env time?

src/presentation/wallet/log-in/index.tsx
```diff
@@ -85,10 +104,7 @@ const LogIn: React.FC = () => {
         <h2 className="text-grayLight text-lg font-medium">
           The ultimate gateway to access the Liquid Network
         </h2>
-        <LogInEnhancedForm dispatch={dispatch} history={history} />
-        {/* <Link className="text-primary block font-bold text-left" to={RESTORE_VAULT_ROUTE}>
```
Janaka-Steph, line 89: This Link should be added back at some point

src/presentation/components/shell-popup.tsx
```diff
@@ -34,13 +37,12 @@ const ShellPopUp: React.FC<Props> = ({
   const goToHome = () => {
     // If already home, refresh state and return balances
     if (history.location.pathname === '/') {
-      dispatch(updateUtxosAssetsBalances(true, refreshCb, (error) => console.log(error)));
```
Janaka-Steph, line 37: Can we restore the refresh functionality?

src/presentation/wallet/log-in/index.tsx
tiero: Yes this is for #116

tiero: @louisinger

louisinger: An issue is enough to remember, no need to keep commented code IMO but I can revert if u prefer

package.json
louisinger: it was just for testing, reverted

Janaka-Steph:
In /presentation/routes/guards.tsx
Module '"../../application/redux/store"' has no exported member 'RootReducerState'.

src/application/utils/transaction.ts
```diff
@@ -140,7 +83,7 @@ export const fillTaxiTx = (
```
Janaka-Steph, line 83: Please add types to fillTaxiTx params.
coinSelector: CoinSelector,
changeAddressGetter: ChangeAddressFromAssetGetter

Janaka-Steph:
I notice the use of .valueOf() in multiple places. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/valueOf
Wondering if it's really necessary

Janaka-Steph:
test/fixtures/wallet.json#L128 
pendingTx block can be removed since it has been removed from wallet state.

src/presentation/wallet/send/choose-fee.tsx
```diff
@@ -1,368 +1,217 @@
...
+  const handlePayFees = (assetHash: string) => {
     if (feeCurrency !== assetHash) {
       setFeeCurrency(assetHash);
       setUnsignedPendingTx('');
-      setSatsPerByte(0);
-      setTopup({});
     }
   };
 
-  /* const warningFee = (
```
Janaka-Steph, line 313: warningFee comment should be added back at some point

src/presentation/wallet/send/choose-fee.tsx
```diff
@@ -1,368 +1,217 @@
...
         setUnsignedPendingTx(tx);
-      } catch (error) {
-        console.log(error);
-        setErrorMessage(error.message);
+        return;
       }
+    } catch (err) {
+      console.error(err);
+      setErrorMessage(`Error for ${assets[feeCurrency].ticker || 'unknow'} asset`);
+      setFeeCurrency(undefined);
```
Janaka-Steph, line 164: In case of error feeCurrency is changed to undefined, which breaks the logo. Would be nice if we can avoid that.

Janaka-Steph:
Testing in Regtest fails at ChooseFee page. 
If sending LBTC asset, clicking on USDT fee causes error "Error for USDt" asset. 
If sending USDT asset, LBTC fee is selected and causes error "Error for LBTC" asset.
changeAddressGetter function is executed twice in a row with a different asset.
When `asset.valueOf() === transaction.sendAsset.valueOf()` is false it should return `transaction.feeChangeAddress?.value` but return undefined. It seems that the value is not ready yet.

src/presentation/wallet/send/choose-fee.tsx
louisinger: [here](https://github.com/louisinger/marina/blob/dee4a8301ffd763a040733cb44a9f3e2b80cedbb/src/presentation/wallet/send/choose-fee.tsx#L204) the `|| ''` should return the default logo path if feeCurrency is undefined

tiero:
There is a regression with pulling down assets name/ticker&precision

image

tiero:
Swaps are not properly recognized

image

louisinger:
UPDATE: #181, always display LBTC balance first in Home screen

tiero:
when switching network we must wipe out the local storage of assets already fetched 

image

PR #189, opened 2021-06-15
Update Marina - LDK v0.3.7

### LINKED ISSUE #176: In Assets list do not hide the row when balance 0, but with previous txs

### LINKED ISSUE #186: The user should be able to "deep restore" its wallet
Use the `gapLimit`' restorer parameter implemented in LDK 0.3.2: https://github.com/vulpemventures/ldk/pull/60

### LINKED ISSUE #187: Change network actions should be done in background
...via web-ext redux aliases

### LINKED ISSUE #188: "Undefined" Bug

louisinger:
**This PR updates the code according to the latest LDK changes, refactors some components, and optimizes the way we persist state.**

- LDK 0.3.3 mainly modifies the way we restore identities. That's why now in wallet state, we have the **last indexes of addresses**.
- Now we restore from indexes, thus we can remove the confidential addresses from the state.
- The transactions are now stored as `TxDisplayInterface`. Because we have never used the large `TxInterface` objects stored in state. Replacing them with `TxDisplayInterface` lets to save disk space and a selector in `transactions.tsx` components (so a faster rendering).
- Refactor the `Select` component.
- add the `Settings/DeepRestorer` page (and the deepRestorer backend logic).

it closes #186 
it closes #176 
it closes #187 
it closes #188 

@tiero @Janaka-Steph please review


tiero:
Stuck here after restore 

image

src/application/redux/reducers/wallet-reducer.ts
```diff
@@ -25,16 +33,27 @@ export function walletReducer(
...
 
-    case ACTION_TYPES.WALLET_SET_ADDRESS_SUCCESS: {
+    case ACTION_TYPES.NEW_CHANGE_ADDRESS_SUCCESS: {
       return {
         ...state,
-        errors: undefined,
-        confidentialAddresses: state.confidentialAddresses.concat([payload.address]),
+        restorerOpts: {
+          ...state.restorerOpts,
+          lastUsedInternalIndex: state.restorerOpts.lastUsedInternalIndex + 1,
```
Janaka-Steph, line 45: If you set it at "+ 1" it is not the last one anymore

src/domain/mnemonic.ts
```diff
@@ -1,19 +1,19 @@
...
   // Trim start-end and replace multiple spaces in between with a single space
   const mnemonic = mnemo.trim().replace(/ +(?= )/g, '');
 
   try {
     // Mnemonic validation, network doesn't matter
-    new MnemonicTdex({
-      // TODO rename
-      chain: 'liquid',
+    new MnemonicIdentity({
+      chain: network,
```
Janaka-Steph, line 13: Does the network matters now? If not we can leave it to 'liquid'

tiero:
> Stuck here after restore
> 
> image

Deleting the cache and reinstalled clean worked again. Noticed is a bit slower than before

src/application/redux/reducers/wallet-reducer.ts
louisinger: NEW_CHANGE_ADDRESS_SUCCESS is dispatched each time we have generated the `nextChangeAddress`, that's why "+1". What do u mean :thinking: ?

src/domain/mnemonic.ts
louisinger: I think we should drop the `new MnemonicIdentity` test and replace it by `bip32.validateMnemonic` function

louisinger:
> > Stuck here after restore
> > image
> 
> Deleting the cache and reinstalled clean worked again. Noticed is a bit slower than before

Yeah the new restorer algorithm is not really optimized, I think we can improve this at the LDK level.

src/domain/mnemonic.ts
```diff
@@ -1,21 +1,13 @@
...
-    new MnemonicIdentity({
-      chain: network,
-      type: IdentityType.Mnemonic,
-      opts: { mnemonic },
-    } as IdentityOpts<MnemonicOpts>);
-    return mnemonic;
-  } catch (err) {
-    throw new Error(err.message);
-  }
+  // Mnemonic validation, network doesn't matter
```
Janaka-Steph, line 10: You can remove the comment

src/presentation/wallet/transactions/index.tsx
```diff
@@ -123,7 +129,11 @@ const TransactionsView: React.FC<TransactionsProps> = ({ assets, transactions, n
             alt="liquid bitcoin logo"
           />
           <p className="text-base font-medium">{txTypeAsString(modalTxDetails?.type)}</p>
-          <p className="text-xs font-light">{modalTxDetails?.blockTime?.format('DD MMMM YYYY')}</p>
+          {modalTxDetails && modalTxDetails.blockTimeMs && (
+            <p className="text-xs font-light">
+              {moment(modalTxDetails.blockTimeMs).format('DD MMMM YYYY')}
```
tiero, line 134: why not `?`

src/domain/mnemonic.ts
```diff
@@ -1,21 +1,12 @@
-import { IdentityOpts, IdentityType, Mnemonic as MnemonicTdex } from 'ldk';
+import { Network } from './network';
+import { validateMnemonic } from 'bip39';
 
 export type Mnemonic = string;
 
-export function createMnemonic(mnemo: string): Mnemonic {
+export function createMnemonic(mnemo: string, network: Network): Mnemonic {
```
tiero, line 6: Is it network needed?

tiero:
some nitpicks, after that LGTM

PR #192, opened 2021-06-30
Implement redux-persist versioning 

### LINKED ISSUE #191: Handle redux-persist schema version
https://github.com/rt2zz/redux-persist/blob/master/src/types.js#L14

louisinger:
**This PR adds the `version` field to persisted reducers.**

- improve `persist` function type and syntax.
- add the `version` field as a parameter in `persist` (initialized to version 1, the current release is 0). Basically, it's a constant number stored in the reducer.
- the migration strategy is the following: if the rehydrated reducer has a version greater than the one defined in scheme, we'll use the `migrate` function defined in `localStorageConfig`.
- next, we should use [MigrationsManifest](https://github.com/rt2zz/redux-persist/blob/master/docs/api.md#createmigratemigrations-config) to implement custom migration for each version number.

it closes #191 

@tiero please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "marina",
-  "version": "0.2.1",
+  "version": "0.2.2",
   "private": true,
   "scripts": {
     "build": "npm run clean && npm run copy-assets && run-p build:**",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "marina",
-  "version": "0.2.2",
+  "version": "0.2.1",
   "private": true,
   "scripts": {
     "build": "npm run clean && npm run copy-assets && run-p build:**",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "marina",
-  "version": "0.2.1",
+  "version": "0.2.2",
```
tiero, line 3: this will be done automatically in the gh action, remove

src/application/redux/reducers/index.ts
```diff
@@ -29,36 +30,90 @@ const browserLocalStorage: Storage = {
...
+    key: 'taxi',
+    version: 1,
+    initialState: taxiInitState,
+  }),
+  connect: persist<ConnectData>({
+    reducer: connectDataReducer,
+    key: 'connect',
+    whitelist: ['enabledSites'],
+    version: 1,
+    initialState: newEmptyConnectData(),
```
tiero, line 115: why not a variable like the others taxiInitState, walletInitState etc..

package.json
tiero: @louisinger

PR #193, opened 2021-07-05
Add `getCoins`, `getBalances`, `getTransactions` and `on` to marina API

### LINKED ISSUE #184: Add `marina.getCoins`, `marina.getBalances` and `marina.getTransactions` and `marina.on` API

### LINKED ISSUE #196: add `isReady` to marina API
isReady should return `true` if the onboarding has been completed

louisinger:
This PR implements the new functions specified [there](https://github.com/vulpemventures/marina-provider/pull/1).

- `getCoins`, `getBalances` and `getTransactions` use redux selectors to return the current state of Marina.
- `marina.on` uses a webext-redux ProxyStore to compare state each time an action is dispatched: and broadcast events according to event's [type](https://github.com/vulpemventures/marina-provider/pull/1/files#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R40).  Then, the injected script add listeners and apply the callback specified by the user in `on` function.

it closes #184 
it closes #196 

@tiero please review


package.json
```diff
@@ -55,7 +55,7 @@
     "ldk": "^0.3.7",
     "lodash.debounce": "^4.0.8",
     "lottie-web": "^5.7.8",
-    "marina-provider": "^1.1.0",
+    "marina-provider": "github:louisinger/marina-provider#enrich-API",
```
louisinger, line 58: need merge of https://github.com/vulpemventures/marina-provider/pull/1

tiero: You can pull from npm `v1.2.0`

src/application/backend.ts
```diff
@@ -382,19 +385,57 @@ export default class Backend {
...
+              } catch (e) {
+                return handleError(id, e);
+              }
+
+            case Marina.prototype.getCoins.name:
+              try {
+                if (!(await this.isCurentSiteEnabled())) {
+                  return handleError(id, new Error('User must authorize the current website'));
+                }
+                const coins = Object.values(marinaStore.getState().wallet.utxoMap);
```
tiero, line 404: Here we get the whole state, after we use selector. Would be better to access state same way

tiero (changes requested):

.github/workflows/main.yml
```diff
@@ -34,19 +34,19 @@ jobs:
           docker-compose -f  tmp/docker-compose.yml up -d
 
       - name: Install dependencies
-        run: npm install --ci
+        run: yarn install
```
tiero, line 37:
```suggestion
        run: yarn install --frozen-lockfile
```

src/application/proxy.ts
```diff
@@ -34,6 +53,39 @@ export default class WindowProxy {
...
+      const payload = (event as CustomEvent).detail;
+      for (const eventListener of this.eventListeners[type]) {
+        eventListener.listener(payload);
+      }
+    });
+  }
+
+  private addEventListener(type: MarinaEventType, listener: MarinaEventListener) {
+    this.eventListeners[type].push(listener);
+    if (this.eventListeners[type].length === 1) this.startWindowListener(type);
```
tiero, line 79: This way we can add a single listener per event type. A developer may want to register many different callbacks for an event

louisinger: no, this line only starts the window's listener the first time we add a callback. but several callbacks are allowed for a same event type here :thinking:

src/presentation/connect/enable.tsx
```diff
@@ -9,24 +9,29 @@ import { useDispatch, useSelector } from 'react-redux';
...
 
   const handleConnect = async () => {
     try {
       await dispatch(enableWebsite(connectData.hostnameSelected, network));
       await dispatch(flushSelectedHostname(network));
       window.close();
+      await windowProxy.proxy('ENABLE_RESPONSE', [true]);
     } catch (e) {
-      console.error(e);
+      window.close();
```
tiero, line 32: close should stay after

tiero:
handle wrong event type name

tiero:
update the readme with yarn instead of npm

src/application/proxy.ts
```diff
@@ -101,3 +106,14 @@ export function makeid(length: number): string {
   }
   return result;
 }
+
+function isMarinaEventType(str: string): str is MarinaEventType {
+  return (
```
tiero, line 111: why not `Object.keys(initialEventListeners)` ? 

it would be hard to maintain all these strings if we add, remove events

src/application/proxy.ts
```diff
@@ -34,6 +53,44 @@ export default class WindowProxy {
       window.location.origin
     );
   }
+
+  on(type: MarinaEventType, callback: (payload: any) => void): EventListenerID {
+    if (!isMarinaEventType(type))
+      throw new Error(
+        'event type is wrong, please choose one of the following: "DISABLED" | "ENABLED" | "NEW_UTXO" | "NEW_TX" | "SPENT_UTXO" | "NETWORK"'
```
tiero, line 60:
```suggestion
       `event type is wrong, please choose one of the following: ${Object.keys(initialEventListeners)} `
```

louisinger:
### Update: Cypress integration

Unfortunately Cypress [does not support browser protocols](https://github.com/cypress-io/cypress/issues/1965) like chrome:extension//. That's why we can't test the Popups UI. However, Cypress is able to install the Marina extension before running tests. 

**Test the API**

The idea is to host a [Marina test app](https://github.com/louisinger/marina-api-test-app/blob/master/README.md), then visit it using Cypress and check if all the tests are running correctly (+ if marina is correctly injected to websites). That's what I've pushed in `cypress/integration` folder. 

Note that I'm building the Marina extension using `NODE_ENV=test` in order to skip the onboarding step and the enabled step (because Cypress can't handle the extension's popups).

`cypress.yml` run `cypress:run` script.

src/application/background-script.ts
```diff
@@ -25,6 +29,15 @@ browser.runtime.onInstalled.addListener(({ reason }) => {
     switch (reason) {
       //On first install, open new tab for onboarding
       case 'install':
+        // /!\ skip onboarding in test env
+        if (process.env.NODE_ENV === 'test') {
+          marinaStore.dispatch(setWalletData(testWalletData));
+          marinaStore.dispatch(enableWebsite('louisinger.github.io', 'regtest')); // skip the enable step too
```
louisinger, line 35: I would host this on a more "private" domain @tiero ?

tiero: wdym by "private"? on VV repos?

tiero: you can transfer the ownership of the repository

tiero:
IsReady should called withour the check if website is enabled, ita used indeed for scenarios like omboarding never been completed

PR #200, opened 2021-07-19
Handle change of the web explorer

### LINKED ISSUE #105: Settings: Handle change of the explorer

louisinger:
This PR modifies the settings/explorer page. 
it adds a `Select` using to modify the "web" explorer = the endpoint redirect while the user is clicking on "see in explorer" buttons. At the moment, two endpoints are supported:
- mempool.space/liquid 
- blockstream.info/liquid 

image

it closes #105 

@tiero please review


tiero:
I would instead assume Esplora-compatible API has the same path deployed

So we should have a select box with Blockstream , Mempool  and Custom.
if user select on Custom them we will show inputs to enter the endpoints manually, both the frontend and backend

The text should be respectively `Esplora URL` and ` Electrs URL`

tiero:
when another option, beside custom is selected, the two inputs should hide.

image

tiero (approved):

PR #217, opened 2021-09-14
Fix Self-transfer tx and `goBack` router crash

### LINKED ISSUE #126: Transactions: Add specific labelling for Self Transfer

louisinger:
TLDR
* Display Self-Transfer tx (as a "deposit" of zero) in tx history + add a `TxType`.
* Fix some buggish behaviors of the router in `SettingsMenu`.

it closes #126 

@tiero please review

### DIFF
--- a/src/presentation/settings/menu-settings.tsx
+++ b/src/presentation/settings/menu-settings.tsx
@@ -6,6 +6,7 @@ import {
   SETTINGS_DEEP_RESTORER_ROUTE,
   SETTINGS_EXPLORER_ROUTE,
   SETTINGS_NETWORKS_ROUTE,
+  DEFAULT_ROUTE
 } from '../routes/constants';
 
 const SettingsMenuSettings: React.FC = () => {
@@ -16,7 +17,11 @@ const SettingsMenuSettings: React.FC = () => {
   const handleDeepRestorer = () => history.push(SETTINGS_DEEP_RESTORER_ROUTE);
 
   return (
-    <ShellPopUp className="h-popupContent" currentPage="Settings">
+    <ShellPopUp
+      className="h-popupContent"
+      currentPage="Settings"
+      backBtnCb={() => history.push(DEFAULT_ROUTE)}
+    >
       <ul className="flex flex-col h-full">
         <button onClick={handleChangeCurrency}>
           <li className="settings-list-item border-t-0.5">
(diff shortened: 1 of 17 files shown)


tiero:
It closes #126 ? If yes add to PR description

src/presentation/settings/menu-settings.tsx
```diff
@@ -16,7 +17,11 @@ const SettingsMenuSettings: React.FC = () => {
   const handleDeepRestorer = () => history.push(SETTINGS_DEEP_RESTORER_ROUTE);
 
   return (
-    <ShellPopUp className="h-popupContent" currentPage="Settings">
+    <ShellPopUp
+      className="h-popupContent"
+      currentPage="Settings"
+      backBtnCb={() => history.push(DEFAULT_ROUTE)}
```
tiero, line 23: why not `goBack`? this will add a new route to the stack

src/presentation/wallet/send/end-of-flow.tsx
```diff
@@ -51,15 +51,15 @@ const EndOfFlow: React.FC<EndOfFlowProps> = ({
...
       });
-    } catch (error) {
+    } catch (error: any) {
       return history.push({
         pathname: SEND_PAYMENT_ERROR_ROUTE,
         state: {
           tx: tx,
-          error: error.message,
-        },
+          error: error?.message || error
```
tiero: If we expect a string (error.message), what are probabilities error is a string already, most likely is an object or undefined? Would be better to JSON.stringify?

src/presentation/wallet/transactions/index.tsx
```diff
@@ -45,10 +46,10 @@ const TransactionsView: React.FC<TransactionsProps> = ({
   const { state } = useLocation<LocationState>();
   const dispatch = useDispatch<ProxyStoreDispatch>();
 
-  const assetImgPath =
+  const getAssetImgPath = () =>
     network === 'regtest'
-      ? imgPathMapRegtest[state.assetTicker] ?? imgPathMapRegtest['']
-      : imgPathMapMainnet[state.assetHash] ?? imgPathMapMainnet[''];
+      ? imgPathMapRegtest[state.assetTicker ?? networks.regtest.assetHash] ?? imgPathMapRegtest['']
+      : imgPathMapMainnet[state.assetHash ?? networks.regtest.assetHash] ?? imgPathMapMainnet[''];
```
tiero, line 52: Shoudldnt be `networks.liquid` ?

src/presentation/settings/menu-settings.tsx
louisinger: to be consistent with the others screens (we also push to home in `Send` and `Receive`). We use  `goback` in all sub-menus (Settings for instance).

PR #223, opened 2021-09-18
Move the provider logic from background to content-script

### LINKED ISSUE #221: Let the broker handles the provider requests via webext-redux
We are running a huge listener aiming to handle the requests of `window.marina` provider in `backend.ts` file. Like we did for `marina.on` method, we should handle the provider requests at the[ broker layer](https://github.com/vulpemventures/marina/blob/f5292116fc4403eaf3d057f1c5516bbd7061c254/src/application/broker.ts?plain=1#L49-L72).

### LINKED ISSUE #222: Sanitize inputs of sendTransaction
If you call `await window.marina.sendTransaction()` without parameters, a popup is shown. This should not be possible and parameter must be validated
﻿

louisinger:
**This PR is a refactor, it deletes, adds, and cleans a lot of things but basically keeps the same architecture. It only drops all the provider logic outside of the background persistent script.**

TL;DR

Firstly I've split the extension scripts into folders (`background`, `content` and `inject`). As they do not share the same JS context, it's easier for code reading to sort files by "context" IMO.

### Background script (`background`)

The background script is now responsible to only (1) [host the background store](https://github.com/vulpemventures/marina/blob/88c2646e0e81dc1524d4c75390de387c7b34524d/src/background/background-script.ts?plain=1#L16) and (2) handle [the creation of the popups](https://github.com/vulpemventures/marina/blob/88c2646e0e81dc1524d4c75390de387c7b34524d/src/background/background-script.ts?plain=1#L79-L85) (via messages from content scripts). No more provider logic there.

### Content script (`content`)

The content script will inject the `inject-script.js` file in web pages and host a broker. The main change is in the broker: it now [handles the messages from web pages via a proxy store](https://github.com/vulpemventures/marina/blob/88c2646e0e81dc1524d4c75390de387c7b34524d/src/content/marinaBroker.ts?plain=1#L97-L240). Basically, we move the former `backend` logic here.

The new changes introduce a [base class](https://github.com/vulpemventures/marina/blob/88c2646e0e81dc1524d4c75390de387c7b34524d/src/content/broker.ts?plain=1#L15) named `Broker`. It lets to abstract the broker logic (message handling, optional proxy store, wait for popup events...) in order to build a custom broker according to our needs. For instance the [MarinaBroker class](https://github.com/vulpemventures/marina/blob/88c2646e0e81dc1524d4c75390de387c7b34524d/src/content/marinaBroker.ts?plain=1#L38) or the [PopupBroker class](https://github.com/vulpemventures/marina/blob/88c2646e0e81dc1524d4c75390de387c7b34524d/src/presentation/connect/popupBroker.ts?plain=1#L9).

### Inject script (`inject`)

Not a lot of major changes. Only a little abstraction of MarinaEvent handler in WindowsProxy.

Tested via Cypress and manual testing.

it closes #221 
it closes #222 

@tiero please review


src/background/backend.ts
```diff
@@ -0,0 +1,285 @@
...
+          break;
+
+        default:
+          break;
+      }
+    });
+
+    browser.alarms.create(UPDATE_ALARM, {
+      when: Date.now(),
+      periodInMinutes: 4,
```
tiero, line 237: lets decrease to 1 minute

PR #227, opened 2021-09-24
Add `ReceiveSelectAsset` page before the address generation

### LINKED ISSUE #226: UX: add additional step in receive button in home to chose the asset
We should re-use the List component used in `Send` button to choose which asset we want to send, but doing the same to choose the asset we want to receive.

We can ass some features asset with the logos, such as L-BTC, USDT, LCAD and add a general purpose "Liquid asset" and additional "Liquid security"

louisinger:
**This PR uses the `AssetListScreen` component from send flow to receive flow in order to add a step before generating wallet addresses**

- abstract the `SelectAsset` screen in order to reuse it in `send-select-asset.tsx` and `receive-select-asset.tsx`.
- add `ButtonList` component (= use the same scrollbar)
- some CSS changes on Home and Transactions views (due to some abstraction in `ButtonList`).
- ATM, the addresses derivation paths do not include the asset type. So all the assets buttons have the same behavior now in `receive-select-asset.tsx`.

it closes #226 

@tiero please review

### DIFF
--- a/src/application/redux/containers/receive-select-asset.container.ts
+++ b/src/application/redux/containers/receive-select-asset.container.ts
@@ -1,19 +1,20 @@
 import { connect } from 'react-redux';
 import { assetGetterFromIAssets } from '../../../domain/assets';
 import { RootReducerState } from '../../../domain/common';
-import SelectAssetView, { SelectAssetProps } from '../../../presentation/wallet/send/select-asset';
+import ReceiveSelectAssetView, {
+  ReceiveSelectAssetProps,
+} from '../../../presentation/wallet/receive/receive-select-asset';
 import { balancesSelector } from '../selectors/balance.selector';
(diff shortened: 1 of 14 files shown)


tiero:
Lint/prettier

tiero (commented):
What do you mean by `ATM, the addresses derivation paths do not include the asset type. So all the assets buttons have the same behavior now in receive-select-asset.tsx.`

louisinger:
> What do you mean by `ATM, the addresses derivation paths do not include the asset type. So all the assets buttons have the same behavior now in receive-select-asset.tsx.`

I mean that theoretically if u click on BTC or USDT u will get the same address. Same behavior in tdex-app, just want to notice that our Mnemonic Identity does not handle the asset type during address generation.

PR #237, opened 2021-10-11
Add Lazy loading, meomized selectors and virtualized scroll on lists

louisinger:
**This PR aims to improve the rendering performance of Marina**

- Lazy loading: lazy load the components at the router levels (React `Suspense` component).  
- Use [react-window](https://github.com/bvaughn/react-window) package in order to render only the components the user see in items lists (transactions, assets and balances lists).
- Use [reselect](https://github.com/reduxjs/reselect) package (memoized selectors). It's basically the same logic in tdex-app.
- Use webpack-bundle-analizer (disabled by default in webpack.prod.js)
- Bonus: some cleaning on the domain layer.

This PR reduces the bundle size (7 MB -> 5MB for index.js).

Empirically, it improves the Marina performance on my system. Please @tiero give me your opinion.


louisinger:
This is no longer relevant since we removed redux (#436)

PR #239, opened 2021-11-30
LDK 0.4 (Testnet support)

### LINKED ISSUE #232: Support Liquid Testnet

louisinger:
- bump LDK to 0.4.0
- use the new types from LDK 0.4.0 -`UtxoInterface` is replaced by `UnblindedOutput`.
- remove the `network` domain, use `NetworkString` type from LDK instead (containing testnet).
- modify views and add explorer URLs according to testnet network type.
- remove the useless "unknow error" when error = undefined in settings pages.

it closes #232 

@tiero please review


tiero (commented):
When switching between network (ie. default liquid to testnet)

The balance are 0 but the list of assets is not updated 

image

tiero (changes requested):
`lbtcAssetByNetwork` util should be updated to support testnet. It is used in `Balance` component to open the esplora URL with the asset

tiero:
If i try to restore indeed, I see that it tries to load from mainnet exploer instead

`https://blockstream.info/liquid/api/address/tlq1qq0thl85nd6x7sp3dhc44h4efssn0cwfv9ac28n0d3u0xd4gcuzyqg2f2ngctfqvcrfl0t858xfwjtdl7mwv9wgx5lug9uqaf0/txs`

tiero (commented):
Address validation on a send page give me invalid for this, generated with Green Testnet wallet

`vjTyyyCeq4AxwBLPv6AR3wYZ27BDDapoiZrPSNnX79MGDgo3iNa6J8ekMjpr3vHv1Co719M3umxcvn9M`

src/presentation/components/shell-popup.tsx
```diff
@@ -93,11 +97,21 @@ const ShellPopUp: React.FC<Props> = ({
...
-            <img className="px-4" src="assets/images/marina-logo.svg" alt="marina logo" />
-          </button>
-          {deepRestorerLoading && <span className="animate-pulse">Deep Restorer loading... </span>}
+        <div className="bg-grayNavBar border-graySuperLight flex flex-row items-center content-center justify-between h-12 border-b-2">
+          <div className="flex flex-row items-center">
+            <button onClick={goToHome}>
+              <img className="px-4" src="assets/images/marina-logo.svg" alt="marina logo" />
+            </button>
+
+            {network === 'testnet' && (
```
tiero, line 106: I woudl add also for `Regtest`

src/presentation/components/shell-popup.tsx
```diff
@@ -121,6 +135,16 @@ const ShellPopUp: React.FC<Props> = ({
       <ModalMenu isOpen={isMenuModalOpen} handleClose={closeMenuModal} />
     </div>
   );
+
+  function getLoaderText(): string | undefined {
+    if (deepRestorerLoading) return 'Deep Restorer loading...';
```
tiero, line 140:
```suggestion
    if (deepRestorerLoading) return 'Restoring...';
```

src/presentation/components/shell-popup.tsx
```diff
@@ -121,6 +135,16 @@ const ShellPopUp: React.FC<Props> = ({
...
+    if (updaterLoaders.txs || updaterLoaders.utxos) return 'Updater loading...';
```
tiero, line 141:
```suggestion
    if (updaterLoaders.txs || updaterLoaders.utxos) return 'Updating...';
```

src/presentation/settings/explorer.tsx
```diff
@@ -14,13 +13,17 @@ import {
...
+import { appInitState } from '../../application/redux/reducers/app-reducer';
 
-function explorerTypesForNetwork(network: Network): ExplorerType[] {
+function explorerTypesForNetwork(network: NetworkString): ExplorerType[] {
   switch (network) {
     case 'liquid':
       return ['Blockstream', 'Mempool', 'Custom'];
     case 'regtest':
       return ['Nigiri', 'Custom'];
+    case 'testnet':
```
tiero, line 25: Would move this after Liquid, so in the UI in Settings we have them sorted

image

tiero:
Add LCAD among featured assets for testnet

Issue #204, opened 2021-07-20
Fetch asset's data in transactions updater


louisinger:

bordalix:
I believe b55af7fd closes this.

PR #238, opened 2021-10-25
Restricted asset account

louisinger:
This PR aims to let the user pair Marina with a cosigner and then "allow" to spend some funds via another injected provider. 

### New Domain: `Account`

- An [Account](https://github.com/vulpemventures/marina/blob/15865412371f2a3b6da042a299e8dedce4326030/src/domain/account.ts#L35-L42) is an interface. The idea is to map **serializable data stored in reducers** to **Accounts**. 
- The `wallet` reducer now stores the account data. Each account is identified by an `AccountID`.
- `encrypted mnemonic` and all the main data used to restore the Main Marina wallet represents the `mainAccount`.
- this "mainAccount" is the single sig account used to own a "regular" asset.
- the user data (utxos, txs etc...) is the **mixed balances of all accounts set up in Marina Wallet**, same for coin selection, blinding, and signing.

### Restricted Asset account

- The user can pair with a cosigner using Settings > Cosigner > Add Cosigner screen. 
- Pairing with a cosigner will add a new account to Marina wallet reducer: RestrictedAssetAccount.
- This account is a 2of2 account working with the `Cosigner` [interface](https://github.com/vulpemventures/marina/blob/03312f936f3554276a73e2d2e912f24d0f221c6a/src/domain/cosigner.ts#L92).
- the user can generates multisig addresses via "Receive" page (:warning: WIP: need to hardcode the assets (with logo and tickers) in select asset screen, now it's a "restricted asset" btn (useful for testing in regtest)).

### "Allowance" provider

the `coinos` provider is injected via `inject-script.ts`. Its implementation is described in [coinosBroker file](https://github.com/vulpemventures/marina/blob/03312f936f3554276a73e2d2e912f24d0f221c6a/src/content/coinos/coinosBroker.ts#L28). It let's to `getCoins` --> returns the utxos owned by the 2of2 Marina account and `allowCoin` --> create a SIGHASH_NONE tx and send it to `Cosigner` (via Cosigner.allow method).

**mocked cosigner**
This PR uses a [mocked cosigner](https://github.com/vulpemventures/marina/blob/03312f936f3554276a73e2d2e912f24d0f221c6a/src/domain/cosigner.ts#L98) for testing purposes as a `Cosigner` interface.

TODO:
This PR is the first base block for the restricted assets but need some task to be fully complete:

- [ ] "real" cosigner implementation.
- [ ] handle the list of restricted assets.
- [ ] Handle allowed coins in Marina ???
[...]

PR #244, opened 2021-12-09
add Account domain and implement redux-saga 

### LINKED ISSUE #242: Invalid balance and tx list on Chrome
- Using Marina commit 108eeed7
- Added Marina extension to Chrome and Firefox (via load unpacked)
- Created two different wallets on each browser (different mnemonics)
- Got 2 L-BTC on Chrome through faucet and sent 1 L-BTC from Chrome to Firefox.
- On the Chrome side, balance is showing 0 (should be 1) and the transaction list shows a value of -2 L-BTC (should be -1).
- On the Firefox side everything is showing properly (balance = 1 and transaction list shows an inbound of 1 L-BTC)
- Esplora shows one of the outputs as non confidential

Left is Chrome, right is Firefox
Screenshot 2021-12-07 at 19 24 43

Screenshot 2021-12-07 at 19 26 07

Steps taken to replicate bug:
1. Create new wallet on Chrome
2. Create new wallet on Firefox
3. nigiri stop --delete
4. nigiri start --liquid
5. generate receiving address on Chrome
6. nigiri faucet --liquid <address from 5>
7. generate receiving address on Chrome
8. nigiri faucet --liquid <address from 7>
9. generate receiving address from Firefox
10. send 1 L-BTC from Chrome to Firefox (to address from 9)
11. ERROR: on Chrome, balance is now 0 (should be 1)
12. ERROR: on Chrome, Tx list shows -2 L-BTC outbound (should be -1)
13. On firefox things appear ok (balance and tx list)
14. Esplora shows one output as non confidential

louisinger:
_This PR is a kind a "rebase" of #238. The idea is to isolate the **domain refacto** + **redux-saga** from 2of2 multisig account logic_ .

**Account domain**

- The idea is to wrap the `IdentityInterfaces` needed to (1) fetch utxos/txs and (2) sign transactions.
- Explained in PR #238

**Redux-saga**

- Redux saga allows replacing two middleware: `redux-thunk` and `background-aliases` (from webext-redux).
- All asynchronous tasks are now managed by "sagas" (including the utxos and txs updater). 
- Thanks to saga, we can implement more complex logic (task cancelation or "worker pattern" for unspent updater).
- I also noticed that saga is much more stable in the web extension context (saga runs in the background script).

**Bug fix**
- fix the first address generation (I discovered that the problem comes from the edge case where lastUsedIndexes are `-Infinity`: now this is handled by the `increment` function in wallet reducer).

it closes #242 
it closes https://github.com/vulpemventures/marina/issues/204

@bordalix @tiero please review


tiero:
Please solve conflicts @louisinger

src/presentation/onboarding/seed-confirm/index.tsx
```diff
@@ -27,7 +27,7 @@ const SeedConfirmView: React.FC<SeedConfirmProps> = ({ onboardingMnemonic, isFro
 
   const handleConfirm = async () => {
     if (selected.join(' ') === mnemonic.join(' ')) {
-      await dispatch(setVerified());
+      await dispatch(setOnboardingVerified());
```
tiero, line 30: can you explain this with a comment

src/presentation/wallet/transactions/index.tsx
```diff
@@ -67,11 +67,6 @@ const TransactionsView: React.FC<TransactionsProps> = ({
     await browser.tabs.create({ url, active: false });
   };
 
-  // Update txs history once at first render
```
tiero, line 70: how do we handle this now?

louisinger: I've reverted the effect in order to dispatch an update task for the MainAccount. Later, if we implement multiple accounts feature we could maybe improve this but now it's ok imo

src/application/redux/containers/choose-fee.container.ts
```diff
@@ -13,7 +13,7 @@ const mapStateToProps = (state: RootReducerState): ChooseFeeProps => ({
   taxiAssets: state.taxi.taxiAssets,
   lbtcAssetHash: lbtcAssetByNetwork(state.app.network),
   sendAddress: state.transaction.sendAddress,
-  changeAddress: state.transaction.changeAddress,
+  changeAddress: state.transaction.changeAddresses[0],
```
tiero, line 16: why we assume is the first element?

src/application/utils/transaction.ts
```diff
@@ -80,6 +80,20 @@ async function blindPset(psetBase64: string, utxos: UnblindedOutput[], outputAdd
   ).toBase64();
 }
 
+function isFullyBlinded(psetBase64: string, excludeAddresses: string[]) {
+  const excludeScripts = excludeAddresses.map((a) => addrLDK.toOutputScript(a));
+  const tx = psetToUnsignedTx(psetBase64);
+  for (const out of tx.outs) {
+    if (out.script.length > 0 && !excludeScripts.includes(out.script)) {
+      if (!out.rangeProof || !out.surjectionProof) {
```
tiero, line 88: not sure this is right check, should be .length > 1 ?

src/presentation/onboarding/seed-confirm/index.tsx
```diff
@@ -6,7 +6,7 @@ import { INITIALIZE_END_OF_FLOW_ROUTE } from '../../routes/constants';
 import Shell from '../../components/shell';
 import { useDispatch } from 'react-redux';
 import { ProxyStoreDispatch } from '../../../application/redux/proxyStore';
-import { setVerified } from '../../../application/redux/actions/wallet';
+import { setOnboardingVerified } from '../../../application/redux/actions/onboarding';
 
 const NULL_ERROR = '';
 const ERROR_MSG = 'Invalid mnemonic';
```
bordalix, line 12: The string 'Invalid mnemonic' is also being used here:
https://github.com/louisinger/marina/blob/account/src/domain/mnemonic.ts#L9

Should we put it in the constants file?

src/application/utils/crypto.ts
```diff
@@ -15,12 +15,16 @@ export function encrypt(payload: Mnemonic, password: Password): EncryptedMnemoni
...
-  return createMnemonic(decrypted);
+  try {
+    const hash = crypto.createHash('sha1').update(password);
+    const secret = hash.digest().slice(0, 16);
+    const key = crypto.createDecipheriv('aes-128-cbc', secret, iv);
+    let decrypted = key.update(encrypted, 'hex', 'utf8');
+    decrypted += key.final('utf8');
+    return createMnemonic(decrypted);
+  } catch {
+    throw new Error('invalid password');
```
bordalix, line 26: On the rest of the code base the string is "Invalid password" (capital I).

And here you're testing the error message against it:
https://github.com/louisinger/marina/blob/account/src/presentation/wallet/send/payment-error.tsx#L48

I think it should be a constant.

src/application/redux/containers/choose-fee.container.ts
louisinger: we always send 1 asset in the popup so it means that we have either 0 or 1 change address (note that `changeAddress` prop can be undefined).

src/application/utils/taxi.ts
```diff
@@ -6,11 +6,11 @@ import {
   TopupWithAssetRequest,
 } from 'taxi-protobuf/generated/js/taxi_pb';
```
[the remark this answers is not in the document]
louisinger: We may track this with an issue on [taxi-protobuf](https://github.com/vulpemventures/taxi-protobuf)

src/application/utils/transaction.ts
louisinger: script for fee outputs is an [emtpy buffer](https://github.com/vulpemventures/liquidjs-lib/blob/master/test/integration/transaction.spec.ts#L204), length > 0 looks good to me.

bordalix (commented):
Console is showing the following warning after login:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at Formik (chrome-extension://edcbiinpejbhkghdkplploabojnaagjo/index.js:41998:19)
    at C (chrome-extension://edcbiinpejbhkghdkplploabojnaagjo/index.js:42377:34)
    at div
    at LogIn (chrome-extension://edcbiinpejbhkghdkplploabojnaagjo/index.js:182832:82)
    at Route (chrome-extension://edcbiinpejbhkghdkplploabojnaagjo/index.js:160652:29)
````

**Comment by bordalix** [comment]
@louisinger if you have 1 L-BTC in your balance, and you try to send 1 L-BTC, without leaving room for the fees,  the extension crashes and it's not possible to get it back. I had to uninstall and install it again. Happens in Chrome and Firefox.

Screenshot 2021-12-14 at 15 05 30

**Review by bordalix — APPROVED** [review]
tACK in Chrome and Firefox.

File: src/domain/account.ts [hunk 8]
```diff
@@ -0,0 +1,82 @@
...
+  restoredMnemonic,
+} from '../application/utils/restorer';
+import { EncryptedMnemonic } from './encrypted-mnemonic';
+import { MasterBlindingKey } from './master-blinding-key';
+import { MasterXPub } from './master-extended-pub';
+
+export const MainAccountID = 'mainAccount';
+export const RestrictedAssetAccountID = 'restrictedAssetAccount';
+
+export type AccountID = 'mainAccount';
```
tiero, line 27:
```suggestion
export type AccountID = MainAccountID;
```

tiero:
Looks like somehow we showing the name of asset instead of the ticker in the list.

image

louisinger:
> Looks like somehow we showing the name of asset instead of the ticker in the list. @louisinger
> 

This is fixed by the last commit. However, u need to reinstall in order to refetch the assets data where ticker = name.

src/application/redux/sagas/updater.ts
```diff
@@ -63,6 +63,7 @@ function* getAddressesFromAccount(account: Account): SagaGenerator<AddressInterf
 
 const putAddTxAction = (accountID: AccountID, network: NetworkString) =>
   function* (tx: TxInterface) {
+    console.log(tx.txid);
```
tiero, line 66: not necessary

tiero:
on SEND_ALL the amount will be filled with `comma`. This seems to be OK-ish with transaction flow, although would be better to user dot as decimal separator always

tiero:
I think the update mechanism is still awkward: I do need to "deep restore" to actually check correct balances.

tiero:
main blocker for me for this change to go in is the "worst" performance for wallet with big utxo set, is arguably much more slower.

Should we try to "rollback" in this PR to the old way to fetch utxos from explorer (instead of new utxosFromTransaction) to prove what is slowing down?

public/manifest.json
```diff
@@ -4,7 +4,6 @@
   "manifest_version": 2,
   "description": "Liquid Wallet Web Extension",
   "permissions": [
-    "alarms",
```
tiero, line 7: why removing this permission?

louisinger: This PR does not use anymore the `alarms` API from the browser (periodic updater is now handled by a specific saga). So Marina doesn't need this permission

tiero: how to periodic updater implements this? setInterval will be killed by the browser runtime.

louisinger: It uses the saga effect [delay](https://github.com/redux-saga/redux-saga/blob/24e9a68d621fd2a57a29b3b80e4b54ecb22fa593/packages/delay-p/src/index.js) (which do not use setInterval but setTimeout on each period).

tiero: Reason why we used alarms in first place

https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#alarms

louisinger: In fact, the periodic updater is only dispatching an action, I don't know if we need a specific service worker for this. Saga seems to perfectly handle timers in the background (it does not cancel the timers) and it offers a way to cancel easily via redux action

src/application/redux/sagas/main.ts
```diff
@@ -61,6 +61,7 @@ function newPeriodicSagaTask(task: () => SagaGenerator, intervalMs: number) {
   return function* (): SagaGenerator<void, void> {
     while (true) {
       yield* task();
+      console.log('periodic task done');
```
tiero: either console.debug or remove it

tiero:
Trying to SEND ALL a non featured asset it crashes and gets the extension stuck to blank page @louisinger 

```js
index.js:2 TypeError: Cannot read properties of undefined (reading 'precision')
    at mapPropsToValues (index.js:2:2977521)
    at i.render (index.js:2:2888228)
    at Ga (index.js:2:2551352)
    at Ha (index.js:2:2551151)
    at Ks (index.js:2:2592096)
    at QA (index.js:2:2578246)
    at wA (index.js:2:2578174)
    at BA (index.js:2:2578037)
    at pA (index.js:2:2575024)
    at index.js:2:2524696
```

tiero:
I have some LCAD in my testnet wallet, in the errors of Marina I get that there was a network error.

Shouldn't we try again in next time we update?

image

louisinger:
> I have some LCAD in my testnet wallet, in the errors of Marina I get that there was a network error.
> 
> Shouldn't we try again in next time we update?

I've added a periodic update used to check each asset of the utxos set every 60s. If the asset is unknown, marina will retry to fetch precision, ticker, and name from the registry.
Send all on "unknown" asset is also fixed.

src/application/redux/sagas/updater.ts
```diff
@@ -245,9 +260,10 @@ export function* watchUpdateTask(): SagaGenerator<void, UpdateTaskAction> {
   }
 
   // start the asset updater
-  const assetsHashChan = yield* createChannel<{ assetHash: string; network: NetworkString }>();
-  yield fork(assetsWorker, assetsHashChan);
-  yield fork(watchForAddUtxoAction, assetsHashChan); // this will fee the assets chan
+  const assetsChan = yield* createChannel<{ assetHash: string; network: NetworkString }>();
+  yield fork(assetsWorker, assetsChan);
+  yield fork(watchForAddUtxoAction, assetsChan); // this will fee the assets chan
+  yield fork(newPeriodicSagaTask(updateUtxoAssets(assetsChan), 60_000)); // retry update of all utxos' assets every 60s
```
tiero, line 266: instead of running this update periodically, we shoudl only do when user refresh.  (ie. click on Marina Logo or goes into transaction tab)

less moving pieces

tiero:
still not clear how we would prevent browser runtime to kill the setInterval that redux-saga (may) use to do periodic updates

src/application/redux/sagas/updater.ts
louisinger: done

public/manifest.json
louisinger: rollback to browser.alarms done

tiero:
Solve conflicts @louisinger

tiero:
Can you solve conflicts and help me in a final Q&A? @bordalix 

I still find occasionally a problem to fetch asset hash for some reason, altough this commit https://github.com/vulpemventures/marina/pull/244/commits/5c4738e1c08973bca950e4b0ec83d1538e707843 should have fixed it (cliecking on marina logo will update also that)

bordalix:
I don't see any conflicts with master, but I do see this bug:

- Click L-BTC then Receive and we have a blank page (bad)
- Click Receive and then L-BTC and we see the QR Code (good)

Could be related: #277

cc/ @louisinger

tiero:
@bordalix should be master already merged here?

bordalix:
> @bordalix should be master already merged here?

It is, was in 2d4aef8 by @louisinger 

I'm putting taxi REST on this to be able to try everything.

bordalix:
Issue: when I change network, it not refresh assets from taxi, so when I try to topup in testnet, it uses `ce0` (USDt in mainnet) in taxi testnet, which obviously fail.

cc @louisinger

tiero:
Please @bordalix  take over on this PR and try to fix where you see fit

tiero:
With latest commit, it hangs (ie. keep loading) as soon I get in the choose fee screen

image

src/application/utils/transaction.ts
```diff
@@ -17,56 +17,127 @@ import {
...
+    return Buffer.from(txid, 'hex').reverse();
+  };
+
+  let index = -1;
+  for (const input of psetToUnsignedTx(pset).ins) {
+    index++;
+    const utxo = utxos.find((u) => txidToBuffer(u.txid).equals(input.hash));
+    if (!utxo) {
+      throw new Error(`blindPSET error: utxo not found '${input.hash.reverse().toString('hex')}'`);
+    }
```
tiero, line 63: There may be occasions where the blinding is either a) not done by us (ccoperative blinding) b) not blinded (like taxi transactions for example)

for a) it's fine to wait new PSETv2 specification that cleanly allows to do this, but for b) we must skip throwing error for unblinded inputs we do not control (ie. provided by taxi)

@louisinger

bordalix: Just a quick note: this code makes impossible to use Taxi, since first input from psetToUnsignedTx is from a utxo the wallet doesn't have (it comes from Taxi), so it will always throw in the case of a Taxi transaction.

I remove it and tested all type of transactions, and everything is working.

Just not sure if by removing it I'm forgetting any edge case. @louisinger ?

louisinger: Indeed throwing an error is not OK in case of unconfidential inputs (taxi especially). However, in case of **confidential input we don't own**, blinding will fail if we are not able to give it the blinders (utxo.blindingData). 
Now, it's not a problem cause we always use Marina's inputs except the taxi ones but they are unconfidential. if in the future marina integrates a feature that uses confidential inputs from another wallet, it will be necessary to modify this function (add a parameter "externalInputsBlindingPrivateKeys" for instance). But I think it's better to wait for PSETv2 for this case as suggested by tiero (which would allow not to manage the blinding keys / blinders exchange).

TL;DR @bordalix no edge cases because Marina never handle cases of confidential inputs owned by another wallet. ACK to remove the check for me :+1:

bordalix:
This happens with me when I'm sending a Tx to blockstream that:
- Is using same inputs sent on previous Tx, or
- Is sending the same Tx (same hash) that was previously accepted

Waiting for 2 minutes should work, since it gives time for previous Tx to confirm and Marina to update list of utxos and Tx.

Having said that, I will raise the HTTP error to the UI.

tiero:
> Waiting for 2 minutes should work, since it gives time for previous Tx to confirm and Marina to update list of utxos and Tx.

was the first tx I made since long time, yes raising the full explorer error would be best

PR #264, opened 2021-12-29
[Hotfix] safe incrementation of address indexes + show default LBTC balance if needed

### LINKED ISSUE #263: First generated address is not persisted in store
The first time we create a Marina wallet and then click on "deposit", the address generated is not persisted to the store.
This is fixed in Account PR (#244) using the [increment](https://github.com/vulpemventures/marina/pull/244/files#diff-9edb817f6c6bb064e14a937ad71b1839ba6e4dd5d4f39aa871c1ff7bf3952c84R188-R192) function.

The user can still receive funds but need to deep restore after first "deposit"

louisinger:
This PR fixes two regressions:
1. on a new wallet, the first address is now correctly persisted (it closes #263)
2. if a wallet doesn't own any LBTC, a default balance is now added (LBTC = 0)

### DIFF
--- a/src/application/redux/selectors/balance.selector.ts
+++ b/src/application/redux/selectors/balance.selector.ts
@@ -1,5 +1,6 @@
 import { balances } from 'ldk';
 import { RootReducerState } from '../../../domain/common';
+import { lbtcAssetByNetwork } from '../../utils';
 
 export type BalancesByAsset = { [assetHash: string]: number };
 /**
@@ -9,5 +10,12 @@ export type BalancesByAsset = { [assetHash: string]: number };
  */
 export function balancesSelector(state: RootReducerState): BalancesByAsset {
   const utxos = Object.values(state.wallet.utxoMap);
-  return balances(utxos);
+  const balancesFromUtxos = balances(utxos);
+
+  const lbtcAssetHash = lbtcAssetByNetwork(state.app.network);
+  if (balancesFromUtxos[lbtcAssetHash] === undefined) {
+    balancesFromUtxos[lbtcAssetHash] = 0;
+  }
+
+  return balancesFromUtxos;
 }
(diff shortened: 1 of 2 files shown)


src/application/redux/selectors/balance.selector.ts
```diff
@@ -9,5 +10,12 @@ export type BalancesByAsset = { [assetHash: string]: number };
  */
 export function balancesSelector(state: RootReducerState): BalancesByAsset {
   const utxos = Object.values(state.wallet.utxoMap);
-  return balances(utxos);
+  const balancesFromUtxos = balances(utxos);
+
+  const lbtcAssetHash = lbtcAssetByNetwork(state.app.network);
+  if (balancesFromUtxos[lbtcAssetHash] === undefined) {
```
tiero, line 16: hasOwnProperty safer?

louisinger: Problem is that balancesFromUtxos is not an object (for Typescript) but a `Record`. However we could use the `in` operator I think but `===` always work as far as I know.

tiero (approved):

src/application/redux/selectors/balance.selector.ts
louisinger: You right but Typescript linter does not allow to use builtin method on "TS type" like `Record`:
```
Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins
```

tiero: https://stackoverflow.com/questions/39282873/how-do-i-access-the-object-prototype-method-in-the-following-logic/39283005#39283005

tiero: not nice syntax, but should be safer

tiero: @Louisinger

tiero:
```suggestion
  if (!Object.prototype.hasOwnProperty.call(balancesFromUtxos, lbtcAssetHash)) {
```

PR #268, opened 2022-01-03
Show value in fractional L-BTC instead of satoshis

### LINKED ISSUE #267: Show values as fractional of L-BTC and not as satoshis
In the follow example, show 0.000085475 instead of 85475:

Screenshot 2021-12-31 at 16 17 07

bordalix:
Fixes #267 

@tiero please review


src/presentation/connect/spend.tsx
```diff
@@ -118,8 +118,9 @@ const ConnectSpend: React.FC<WithConnectDataProps> = ({ connectData }) => {
           <div className="h-64 mt-4 overflow-y-auto">
             {connectData.tx?.recipients?.map((recipient: RecipientInterface, index) => (
               <div key={index}>
+                {console.log('xxxxx', recipient)}
```
tiero, line 121: remove

PR #280, opened 2022-01-11
Fixes back from Receive

### LINKED ISSUE #277: Bug after QR code to receive
There are 2 bugs actually.

How to replicate:
1. Fresh install of the Chrome extension
2. Restore wallet (don't need to save mnemonic)
3. Click on LBTC
4. Click on Receive
5. Asks to save mnemonic (bug 1)
6. Click Save and it goes to QR code
7. Click back and you get a blank page (bug 2)

The component TransactionsView (called after step 7) has no `state` defined, so it blows.

If you click directly on Receive on the home screen, there's no issue:
1. Fresh install of the Chrome extension
2. Restore wallet
3. Click on Receive
4. Click Save and it goes to QR code
5. Click back and you go to the home screen

bordalix:
Fixes #277

This PR fixes two issues:
1. When going back from Receive the app would crash (see #277 for details)
2. When clicking on Receive from the Asset view the app would always say the mnemonic was not saved
3. Clicking save on the Save Mnemonic modal from step 2 would proceed to Address view

isWalletVerified is now passed as props to Transactions view.

@tiero please review

### DIFF
--- a/src/presentation/wallet/receive/index.tsx
+++ b/src/presentation/wallet/receive/index.tsx
@@ -9,6 +9,7 @@ import { updateUtxos } from '../../../application/redux/actions/utxos';
 import { ProxyStoreDispatch } from '../../../application/redux/proxyStore';
 import { masterPubKeyRestorerFromState, MasterPublicKey, StateRestorerOpts } from 'ldk';
 import { incrementAddressIndex } from '../../../application/redux/actions/wallet';
+import { DEFAULT_ROUTE } from '../../routes/constants';
 
 export interface ReceiveProps {
   pubKey: MasterPublicKey;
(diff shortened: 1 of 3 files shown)


src/presentation/wallet/receive/index.tsx
```diff
@@ -23,7 +24,7 @@ const ReceiveView: React.FC<ReceiveProps> = ({ pubKey, restorerOpts }) => {
   const [buttonText, setButtonText] = useState('Copy');
   const [isAddressExpanded, setAddressExpanded] = useState(false);
   const handleExpand = () => setAddressExpanded(true);
-  const handleBackBtn = () => history.goBack();
+  const handleBackBtn = () => history.push(DEFAULT_ROUTE);
```
tiero, line 27: maybe replace is more appropriate here?

bordalix: Yes, you're right.

tiero:
tACK Chrome

PR #287, opened 2022-01-12
fix icon when sending full balance

### LINKED ISSUE #286: Sutract fee when send amount equals to current LBTC balance
If I try to send my total balance, an error is thrown (due to lack of funds for paying the fee), the catch() undefines `feeCurrency` so I lose the LBTC icon on the `Choose fee` component. See video:

https://user-images.githubusercontent.com/9318412/149127177-ceb5e7df-aae5-41ea-a19d-03871684f170.mov

I don't see the reason to undefine `feeCurrency`:

```
    } catch (err) {
      console.error(err);
      setErrorMessage((err as Error).message);
      setFeeCurrency(undefined); // Here
      setFeeChange(undefined);
      setTopup(undefined);
    }
```
in `src/presentation/wallet/send/choose-fee.tsx`

bordalix:
Fixes #286 

@tiero please review

### DIFF
--- a/src/presentation/wallet/send/choose-fee.tsx
+++ b/src/presentation/wallet/send/choose-fee.tsx
@@ -108,7 +108,11 @@ const ChooseFeeView: React.FC<ChooseFeeProps> = ({
     } catch (err) {
       console.error(err);
       setErrorMessage((err as Error).message);
-      setFeeCurrency(undefined);
+      // undefining feeCurreny will make the component lose the currency icon.
+      // this error happens if the user tries to send the full balance.
+      // see https://github.com/vulpemventures/marina/issues/286 for more details
+      // commenting the next line to prevent this
+      // setFeeCurrency(undefined);
       setFeeChange(undefined);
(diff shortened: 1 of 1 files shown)


tiero:
do you recall the reason for that? @louisinger  

Also how this will change with "send all" feature in the #244 PR?

louisinger:
> do you recall the reason for that? @louisinger  

The idea was to set fee currency as "not selected" (undefined) if an error happens. It forces the user to click again in order to re-trigger taxi topup call or fee estimation (it depends on the fee asset)

> Also how this will change with "send all" feature in the #244 PR?

#244 should not throw an error while "send all", so it should work there.

tiero:
I think is fine the current behavior: uplifting the error to the user and prevent him to select the fee and therefore proceed.

What actually we should do here is instead to pass to a "subtract fee amount" scenario, like the SEND ALL case of #244

bordalix:
Agree. Closing this PR.

PR #292, opened 2022-01-20
Rest in taxi

### LINKED ISSUE #290: Taxi connection is not working
If I try to send USDT in the "Choose fee" page I should see the option to select Taxi, which is not the case.

If I call ListAssets on grpc.liquid.taxi with Bloom it works

image

bordalix:
Fixes #290 

Access taxi services via REST (removes gRPC).
Remove `taxi-protobuf` dependency.
Throw response (and not response.data) on broadcastTx.

@tiero please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "crypto-browserify": "^3.12.0",
     "decimal.js": "^10.2.1",
     "formik": "^2.2.6",
-    "google-protobuf": "^3.15.8",
     "ldk": "^0.4.0",
     "lodash.debounce": "^4.0.8",
     "lottie-web": "^5.7.8",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "redux-thunk": "^2.3.0",
     "stream-browserify": "^3.0.0",
     "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.2",
-    "taxi-protobuf": "vulpemventures/taxi-protobuf",
     "uuid": "^8.3.2",
     "webext-redux": "^2.1.7",
     "webextension-polyfill": "^0.8.0",
--- a/src/application/utils/taxi.ts
+++ b/src/application/utils/taxi.ts
@@ -1,24 +1,24 @@
 import axios from 'axios';
 
 interface AssetDetails {
-  assetHash: string,
-  assetPrice: number,
-  basisPoint: number,
+  assetHash: string;
+  assetPrice: number;
+  basisPoint: number;
 }
 
 export interface Topup {
-  assetAmount: number,
-  assetHash: string,
-  assetSpread: number,
-  partial: string,
-  topupId: string,
+  assetAmount: number;
+  assetHash: string;
+  assetSpread: number;
+  partial: string;
+  topupId: string;
 }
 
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -50,7 +50,6 @@
     "redux-thunk": "^2.3.0",
     "stream-browserify": "^3.0.0",
     "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.2",
-    "taxi-protobuf": "vulpemventures/taxi-protobuf",
```
tiero, line 53: looks like we could remove google-protobuf too?

tiero:
testing on Chrome I am getting `gRPC requires HTTP/2`. This most likely beacuse you are not passing content type `application/json`. We multiplex on the same port many protocols and switch based on content type.

src/application/utils/taxi.ts
```diff
@@ -1,24 +1,35 @@
...
-  asset: string
-): Promise<TopupWithAssetReply.AsObject> => {
-  const client = new TaxiClient(taxiUrl, undefined);
-  const request = new TopupWithAssetRequest();
-  request.setAssetHash(asset);
-  const res = await client.topupWithAsset(request, null);
-  return res.toObject();
+  assetHash: string
+): Promise<TopupWithAssetReply> => {
+  const { data } = await axios.post(`${taxiUrl}/asset/topup`, assetHash);
```
tiero, line 33: We miss third parameter, with an object and a key `headers` where we specify application/json

strange works for you

bordalix: It was working because I had `{ assetHash }` has second parameter, and when we have an object has second parameter axios adds the `content-type application/json` by default. My last commit introduced the error, was not properly tested. This new commit 9eaef50 fixes it. @tiero can you try it?

tiero:
I tried to send 1USDt  but seems it multiples wrongly to be used for coin selection

```
Error: Not enought coins to select 10000000049500000 f3d1ec678811398cd2ae277cbe3849c6f6dbd72c74bc542f7c4b11ff0e820958 (has: 316666666)
```
Screenshot 2022-01-21 at 11 08 50
Screenshot 2022-01-21 at 11 08 57

bordalix:
> I tried to send 1USDt but seems it multiples wrongly to be used for coin selection
> 
> ```
> Error: Not enought coins to select 10000000049500000 f3d1ec678811398cd2ae277cbe3849c6f6dbd72c74bc542f7c4b11ff0e820958 (has: 316666666)
> ```

Yes, yesterday I notice this also, but it seems to happen on `master` also (can you confirm?) so it seems not related with this PR. Also, I have an idea this is addressed on @louisinger PR.

package.json
bordalix: Commit 119c83a (just now) removes google-protobuf

tiero:
werid since I can't test the taxi topup then

tiero:
> but it seems to happen on `master` also (can you confirm?) 

master seems to  have outdate taxi-protobuf, therefore cant ListAssets correctly and test taxi coin selection

bordalix:
I'm working with @louisinger to have PR #244 stable (and with Taxi REST) in order for us to be able to test it and finally merge it.

PR #299, opened 2022-02-01
raises post to explora error to UI

bordalix:
@tiero please review

### DIFF
--- a/src/application/utils/network.ts
+++ b/src/application/utils/network.ts
@@ -1,14 +1,24 @@
-import axios from 'axios';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import axios, { AxiosError } from 'axios';
 import { networks, NetworkString } from 'ldk';
 import { IAssets } from '../../domain/assets';
+import { extractErrorMessage } from '../../presentation/utils/error';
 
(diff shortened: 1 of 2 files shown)


src/application/utils/network.ts
```diff
@@ -1,14 +1,24 @@
...
-    throw new Error(response.data);
+  try {
+    const response = await axios.post(`${baseUrl}/tx`, txHex);
+    if (response.status !== 200) {
+      console.error(response.data);
+      throw new Error(response.data);
+    }
+    return response.data;
+  } catch (error: unknown | AxiosError) {
+    console.log(error);
```
tiero, line 16: I would avoid log errors, since we throw up already

src/application/utils/network.ts
```diff
@@ -1,14 +1,24 @@
...
 
 export const broadcastTx = async (baseUrl: string, txHex: string): Promise<string> => {
-  const response = await axios.post(`${baseUrl}/tx`, txHex);
-  if (response.status !== 200) {
-    console.error(response.data);
-    throw new Error(response.data);
+  try {
+    const response = await axios.post(`${baseUrl}/tx`, txHex);
+    if (response.status !== 200) {
+      console.error(response.data);
```
tiero, line 11: remove

PR #303, opened 2022-02-07
Validate recipient object

### LINKED ISSUE #302: sendTransaction: Recipient object must be validated
If you try to use `window.marina.sendTransaction` with a non compliant Recipient object without `asset` it will not block you and let proceed till coin selection.

```
await marina.sendTransaction([{address: "tex1qhdvpxjjjumvms8qhdwfcmjg27ww0xr5nsf406f", value: 5000000}]); 
```

It should throw error early.

We should check:

1) address & value being present ALWAYS.
2) If asset is not present either default to L-BTC of the selected network since is known 
3) address must match and be valid for the current network.

bordalix:
Fixes #302

This PR:
- validates asset and value on all recipients in sendTransaction()
- moves value validation from provider.ts to marinaBroker.ts

@tiero please review

### DIFF
--- a/src/content/marina/marinaBroker.ts
+++ b/src/content/marina/marinaBroker.ts
@@ -178,8 +178,20 @@ export default class MarinaBroker extends Broker {
           const lbtc = lbtcAssetByNetwork(selectNetwork(state));
           const feeAsset = feeAssetHash ? feeAssetHash : lbtc;
 
+          // validate if fee asset is valid
           if (![lbtc, ...selectTaxiAssets(state)].includes(feeAsset)) {
             throw new Error(`${feeAsset} not supported as fee asset.`);
+          }
+
+          // validate object recipient (asset and value)
+          // if no asset is present, assume lbtc for the current network
+          for (const recipient of recipients) {
+            if (!recipient.asset) {
+              if (!lbtc) throw new Error('missing asset on recipient');
+              recipient.asset = lbtc;
+            }
+            if (!recipient.value) throw new Error('missing value on recipient');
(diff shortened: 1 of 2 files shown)


src/content/marina/marinaBroker.ts
```diff
@@ -178,10 +178,22 @@ export default class MarinaBroker extends Broker {
...
 
+          // validate object recipient (asset and value)
+          // if no asset is present, assume lbtc for the current network
+          for (const recipient of recipients) {
+            if (!recipient.asset) {
+              if (!lbtc) throw new Error('missing asset on recipient');
+              recipient.asset = lbtc;
+            }
+            if (!recipient.value) throw new Error('missing value on recipient');
+            if (recipient.value < 0) throw new Error('negative value on recipient');
```
tiero, line 194: worth checking floating point numbers as not valid

tiero (approved):

src/content/marina/marinaBroker.ts
bordalix: Adding Number.isSafeInteger in 277de1f

But, `Number.isSafeInteger(1.0)` is true, which means a value of `1.0` would continue downstream.
Should I `recipient.value = parseInt(recipient.value, 10)` after this test?

tiero: yes add extra parseInt indeed

tiero: good catch

tiero (changes requested):
parseInt after isSafeInteger

PR #309, opened 2022-02-11
Fix input blinding data

bordalix:
This PR:

- fixes error detecting if an input is confidential (unblindData is defined in all inputs after all)
- fixes error that occurred when two inputs were from the same Tx (but with different vouts)

@louisinger @tiero please review

### DIFF
--- a/src/application/utils/transaction.ts
+++ b/src/application/utils/transaction.ts
@@ -18,6 +18,7 @@ import {
   getAsset,
   NetworkString,
   IdentityInterface,
+  isConfidentialOutput,
 } from 'ldk';
 import { confidential, networks, payments, Psbt } from 'liquidjs-lib';
 import { isConfidentialAddress } from './address';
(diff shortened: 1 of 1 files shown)


louisinger (approved):
tACK

bordalix:
Why `secp256k1_surjectionproof_initialize` was so random?

The error would only throw when there were 2 inputs coming from the same transaction.

The inputs would have the same hash (but different vouts), and on the second run of the cycle, utxos.find() would return the first utxo (again), instead of the second, thus generating the error:

```
const utxo = utxos.find((u) => txidToBuffer(u.txid).equals(input.hash));
```

By adding an additional condition to the find, we rest assure the utxo found is the correct one:

```
const utxo = utxos.find(
      (u) => txidToBuffer(u.txid).equals(input.hash) && u.vout === input.index
);
```

This happend with bitmatrix because when we swap LBTC for USDt, they normally send us 2 coins, from the same Tx, with half the value each. Every time this 2 coins were selected for a new Tx, we would have the error thrown.

tiero (approved):

PR #313, opened 2022-02-21
Fix alarms

bordalix:
**Update:** after several iterations, this PR context changed. See [new comment](https://github.com/vulpemventures/marina/pull/313#issuecomment-1050786927)

**Oudated:**

I made some changes on the way Marina synchronises utxos, txs and taxi assets that I think will make us better net citizens.

Today, Marina is hammering explorers (and taxi) every minute with hundreds of requests, which can be cumbersome and even black list IP addresses. This happens from the minute Marina is installed and is kept forever.
So I decided to try a new approach: now Marina will only try to synchronise when the user is authenticated. After the 500 seconds idle timeout, the user will be logged out and Marina will stop syncing and only restart again after a successful login.

In a nutshell, login => start syncing, logout => stop syncing.

Also, previously Marina would start the periodic updater, but it would only make the first sync one minute after. So I changed it to sync immediately after the creation of the periodic update task, so a successful login will trigger an immediate sync.

Also,  reloadCoins() is now a util function (in marinaBroker.ks) so we can call it whenever we see fit. For instance, every time  someone calls getCoins(), it will first reloadCoins() and only after will return the available coins. This behaviour was implemented in the following functions:

- sendTransaction()
- getTransactions()
- getCoins()
- getBalances() 

Finally, there was a bug when getting the assetHash to query the explorer, fixed in 08dd949

@louisinger @tiero please review.


tiero:
As I said in offline chat, Marina users are not only "humans" ie. interacting with Popup via UI, but especially other web apps that interact programmatically. Stopping updatings utxos if user is not "logged in" will not catch new inbound/outbound txs: think about a web app that shows an address via `getNextAddress`, has others send to it and/or then spend funds via `signPset` etc.. we will be with a not updated utxo set.

tiero:
re: reloadCoins

This is a workaround, but not optimal in terms of speed:
The pattern of "I keep watching the chain so when you need to select coin to spend, they will be already cached" is done to speedup transaction creation, for the time being is fine: better being slower than having outdated view of the utxo set.

src/background/alarms.ts
```diff
@@ -44,4 +53,9 @@ export const periodicUpdater = newPeriodicTask(
   dispatchUpdateTaskForAllAccountsIDs,
   1
 );
+
 export const periodicTaxiUpdater = newPeriodicTask(Alarm.TaxiUpdate, dispatchUpdateTaxiAction, 1);
+
+export const clearAllPeriodicUpdaters = () => {
```
tiero, line 59: I am bit of nACK in stopping updading utxo set after logout, for reason stated in main comment

src/content/marina/marinaBroker.ts
```diff
@@ -100,9 +102,27 @@ export default class MarinaBroker extends Broker {
...
   }
 
+  private async reloadCoins(store: BrokerProxyStore) {
+    const state = store.getState();
+    const network = selectNetwork(state);
+    const allAccountsIds = selectAllAccountsIDs(state);
+    const makeUpdateTaskForId = (id: AccountID) => updateTaskAction(id, network);
+    allAccountsIds.map(makeUpdateTaskForId).map((x: any) => store.dispatch(x));
+    // wait max 10 seconds for update to finish
+    let attempts = 10;
```
tiero, line 112: not sure 10 seconds could be enough for "deep" wallets to complete: what would happen in that case?

the essence of this API is to "load" the latest state possible of the utxo from blockchain, no matter how much time it takes.

bordalix:
> As I said in offline chat, Marina users are not only "humans" ie. interacting with Popup via UI, but especially other web apps that interact programmatically. Stopping updatings utxos if user is not "logged in" will not catch new inbound/outbound txs: think about a web app that shows an address via `getNextAddress`, has others send to it and/or then spend funds via `signPset` etc.. we will be with a not updated utxo set.

That's why I force a reloadCoins() before any action that uses utxos.
Having said that, I could be missing something.
I will double check.

src/content/marina/marinaBroker.ts
bordalix: Good question.

It would load the same, but the request (ex: getCoins()) could be returning from a previous utxo set.

Spent days trying to figure how to force wait for reloadCoins() before proceeding. With the help of @louisinger this was the possible solution. Not happy about this solution, but not sure on how to implement it any other way.

bordalix: I tried with
```
while (store.getState().wallet.updaterLoaders !== 0) {}
```
but this becomes an infinite loop, since the event loop is never freed to perform the requires updateTask and so updaterLoaders never gets to 0.

tiero: I do think reloadCoins is just suboptimal solution, which is the reason we do background updates in the first place: no need to call calls in sync when needed, but just get from storage

louisinger: I think the way we implement loaders is limited for that case: currently, `updaterLoaders` is an integer and let us know if there are **one or more** updaters running. To wait for a specific task properly, I think it misses two pieces:
* identity each UPDATE_TASK by an unique ID 
* associated a loader to each updater ID in the marina redux state (replace `updaterLoaders`)

something like:
```js
const id = marina.reloadCoins()
while (selectUpdaterIsLoading(id)(marinaStore.getState())) {
   // wait....
}
// done 
```

tiero: Agreed, toughts @bordalix ?

bordalix: Agree, 5ef5337 is going into that direction.

src/application/redux/sagas/updater.ts
```diff
@@ -153,6 +158,7 @@ function* txsUpdater(
     txsGenenerator,
     putAddTxAction(accountID, network, walletScripts)
   );
+  console.log(`${new Date()} txs updated`, txsHistory);
```
tiero, line 161: we can remove this

src/application/utils/taxi.ts
```diff
@@ -24,6 +24,7 @@ export interface TopupWithAssetReply {
 
 export const fetchAssetsFromTaxi = async (taxiUrl: string): Promise<string[]> => {
   const { data } = await axios.get(`${taxiUrl}/assets`);
+  console.log(`${new Date()} taxi updated`, data);
```
tiero, line 27: remove now

tiero:
Can you update the PR description with actual changes happended for future readers?

bordalix:
After some iterations, this PR was reduced to:
- Bug fix where explorer URL for a given asset was broken
- Removal of setIdleAction(), since same functionality is already implemented
- Update utxos, transactions and taxi on first run (don’t wait one minute to do it)
- Wait for reloadCoins() to finish (aka blocking) before returning
- Add console.log() of received utxos from ldk for easier debugging
- Re-enable background syncs every time (and not only when authenticated)
- reloadCoins() is now a private function in src/content/marina/marinaBroker.ts
- Bug fix where state was not updated after reloadCoins()

src/application/redux/sagas/updater.ts
```diff
@@ -99,6 +102,7 @@ function* utxosUpdater(
   for (const utxo of toDelete) {
     yield* putDeleteUtxoAction(accountID, network)(utxo);
   }
+  console.log(`${new Date()} utxos received`, receivedUtxos);
```
tiero, line 105:
```suggestion
  console.debug(`${new Date()} utxos received`, receivedUtxos);
```

PR #317, opened 2022-03-02
Lock utxos

### LINKED ISSUE #316: Locking utxos
After used in a transaction, utxos should be locked to prevent:
- double spending issues
- bloated balance (when change utxo is added to the set)

bordalix:
Fixes #316

This PR:
- Creates a new action LOCK_UTXO
- Adds lockedUtxos to state
- Lock utxos after successful broadcast [see rational]
- Coin selection avoids filtered utxos
- Utxos are unlocked after 5 minutes when next ADD_UTXO
- Edge case of payment error and retry covered

Rational:
- Initial idea was to lock utxos on coin selection, but since this happens before the user’s final decision (Accept/Retry) we cannot lock utxos on that stage. We must only lock the utxos after a successful broadcast.
- This means that on send and spend, we need to know the selected utxos used for the transaction, and lock them after successful broadcast. This means createSendPset() and createTaxiTxFromTopup() need to return the selected utxos used in the pset to makeTransaction()
- Available coins for greedyCoinSelection() are filtered beforehand, eliminating all locked utxos. This is done via the function selectUtxos().
- Utxos are unlocked when a new coin arrives, but only if they are locked for more than 5 minutes. This is done via the action ADD_UTXO.
- For code simplicity, the lockedUtxos are placed at state.wallet.lockedUtxos and not inside the account. Trusting on collision resistance allows for a simpler code.

@tiero please review


src/presentation/wallet/send/payment-error.tsx
```diff
@@ -7,32 +7,39 @@ import {
...
-        history.push({
-          pathname: SEND_PAYMENT_SUCCESS_ROUTE,
-          state: { txid },
-        });
-      })
-      .catch((error) => {
-        console.error(error.message);
-      });
+  const handleRetry = () => {
+    const txid = broadcastTx(explorer, state.tx);
```
tiero, line 31: await?

bordalix: fixed with b16d01c

tiero:
I do think "crediting" the change output could be a nice UX for the end user: ideally this utxo should be market as if it was locked (ie. cant be used as utxo for a new tx until 1 confirmation)

Ideally when we fetch from explorer an outpoint that matched our virtual utxo, then we can "delete" it and keep the actual one via ADD_UTXO action

What do you think?

bordalix:
Commit 34dcd98 creates an unconfirmed utxo from the change and adds it immediately to the utxo set for selection.

This allows:
- No more balances going to 0 when the transaction used all coins
- Being able to use change utxo immediately without having to wait for confirmation

For each change address, we hand craft the utxo with:
- The txid
    - Given by the result of broadcastTx()
- The vout
    - Get the transaction from the signed tx hex
        - transaction = Transaction.fromHex(signedTxHex)
    - Calculate changeOutputScript for the change address
        - changeOutputScript = address.toOutputScript(address, network);
    - Find index in transaction.outs where script = outputScript
        - vout = transaction.outs.findIndex(...)
- The prevout
    - Equal to transaction.outs[vout]
- The unblindData
    - Get the blinding private key from the Identity Interface 
        - blindPrivKey =identities[0].getBlindingPrivateKey(prevout.script)
    - Unblind data
        - utxo = unblindOutput({ txid, vout, prevout }, blindPrivKey);

After this, the utxo is added to the utxosMap on wallet state and can be immediately selected.

src/application/redux/reducers/wallet-reducer.ts
```diff
@@ -28,13 +28,28 @@ export const walletInitState: WalletState = {
   },
   updaterLoaders: 0,
   isVerified: false,
+  lockedUtxos: {},
+};
+
+// returns only utxos locked for less than 5 minutes
+const filterOnlyRecentLockedUtxos = (state: WalletState) => {
```
louisinger, line 35: I think u can simplify and optimize this using a redux-saga (tdex-app [example](https://github.com/tdex-network/tdex-app/blob/ce03725b0d4a82f6d0092caa1eeea3baa4fa9520/src/redux/sagas/walletSaga.ts#L106-L110)) instead of a filter function. Here, it looks like the LockedUtxos are only "unlocked" if an unspent is added (where the filter function is called).

src/application/redux/reducers/wallet-reducer.ts
```diff
@@ -176,6 +220,22 @@ export function walletReducer(
...
+        lockedUtxos: {
+          ...state.lockedUtxos,
+          [toStringOutpoint(utxo)]: Date.now(),
+        },
+      };
+    }
+
+    case ACTION_TYPES.ADD_UNCONFIRMED_UTXOS: {
+      const { unconfirmedUtxos, accountID, network } = payload;
+      return addUnconfirmed(state)(unconfirmedUtxos, accountID, network);
```
louisinger, line 236: I think we should handle the case where the "unconfirmed utxo" is never confirmed (= the updater does not return the "real" coin). If the previous tx is not confirmed for any reasons, our "fake" coin could cause missing-inputs error later. 

As a solution, we could remove the unconf coin if not confirmed after 1 min for instance?

tiero (approved):
tACK on Chrome

src/application/redux/reducers/wallet-reducer.ts
tiero: A broacasted TX can indeed being "replaced" by new txid indeed (replace by fee), so the change output can be different, so yes current virtual output can change.

Maybe we can move this it in a specific ticket and address it later?

src/application/redux/reducers/wallet-reducer.ts
```diff
@@ -28,13 +28,28 @@ export const walletInitState: WalletState = {
   },
   updaterLoaders: 0,
   isVerified: false,
+  lockedUtxos: {},
+};
+
+// returns only utxos locked for less than 5 minutes
+const filterOnlyRecentLockedUtxos = (state: WalletState) => {
+  const expiredTime = Date.now() - 300_000; // 5 minutes
```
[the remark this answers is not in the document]
tiero: moreover if this utxo is actually spen by real tx onchain, we may want to remove it lock earlier? Although this is safe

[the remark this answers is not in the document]
bordalix: In testnet I've seen the explorer returning already spent utxos 2 or 3 minutes later. It's wrong, it shouldn't happen, but I've seen it. The choice of 5 minutes was more on the safe side.

src/application/redux/reducers/wallet-reducer.ts
tiero: Agree with @bordalix

tiero: Shall we move this in a redux saga?

> it looks like the LockedUtxos are only "unlocked" if an unspent is added
@louisinger  what would be the problem of not unlocking? As soon a transaction is submitted to mempool, can't be canceled (unless replace by fee, but at this point the new utxo state will be synced from network)

tiero: we agreed offline with @bordalix to address in this PR only the Louis concern, but not about the no-network issue (can be done later on in a dedicated PR with proper network handling now is missing)

bordalix:
Let's imagine this:
- Steve sends to Bob
- Utxos are locked for 5 minutes
- For any reasons the tx is not broadcasted (so the locked utxos are not spent)
- 5 mins later, tx is not broadcasted so utxos are unspents but locked (because steve does not receive any coins during the locked time, so the filter function is never applied).
- To be able to spend the "unspent locked coin" after an error, Steve has to receive a coin, which can take forever

The solution is running the filter that unlocks utxos not on addUnspent(), but on **every successful update**. On other hand, the filter should **ONLY** run on successful updates, in order to assure disconnected wallets will still behave doing coin selection after 5 minutes.

Commit 848990a does this.

Issue #250, opened 2021-12-18
Support multiple script types


tiero:
preliminary read: https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md

In the eventuality to support custom scripts, beside "pay to pubkey" and "pay to multisig", having Marina supporting[ loading descriptors](https://github.com/vulpemventures/marina-provider/issues/11) would benefit developers of more advanced smart contracts/covenants, to have Marina to track custom script types.

At the moment we are [refactoring](https://github.com/vulpemventures/marina/pull/244) the domain to support many "accounts" which is a good first step, but we should abstract away the "bag of keys" model and embrace the output descriptor way, which means we are not constrained by having an account linked to a derivation path/extended public key (or collection of it in case of multisig)

## Current model

The new model will be as follow:

- We have a secret (mnemonic), we derive signing and blinding keys
- We have "Main Account"  with a default template ie. `wpkh(<xpriv|xpub>)`
- We have "2of2 Account" with a default template ie. `wsh(sortedmulti(2,<marina_xpriv>,<cosigner_xpub>))` (maybe more complicated with timelock stuff)
- We have N account(s) loaded arbitrary with descriptors we may want to keep track of. More complex scenario will be handled with `raw(SCRIPT)`

This will allow to keep track of balances and show in UI

## Importing descriptors

The loading of descriptor need to create extra metadata about `hasprivatekeys` when loaded with keys via `xpriv` (need to think about this if want to do thi, how we handle this keys etcc.) and `ownkeys` when we should use Marina's mnemonic to generate keys.

A suggestion can  be to have a special reserved keyword called `marina` so we can do things like `wsh(sortedmulti(2,marina,<cosigner_xpub>))` and we can fill there the marina xpub 

## Signing 

The signing get a bit tricky, since we need to have a way to link a script to a specific key, depending how we imported this. 

## Restore

This is tricky: since we can't know all the possible descriptors/complex raw scripts, user must backup also those and re-load after we restore from mnemonic later on. So dedicated API and/or UI must be designed.

PS: This is def something that could belong to `ldk` itself, but for recent experience, starting developing abstractions without being on the "ground" ie. wallet/ux development, would be much harder, so suggest to start desiging/modeling here and then port there.

tiero:
Concrete use case: 

bitmatrix.app wants to create this kind of addresses:

```OP_1 OP_CHECKSEQUENCEVERIFY <PubKey> OP_CHECKSIG``` 

[...]

PR #325, opened 2022-03-15
Marina output descriptors

### LINKED ISSUE #320: Implement script type `Template`
related to #250

louisinger:
_This PR does not add any "real" features to Marina, `descriptors` is a #250 preliminary work._

TL; DR:
* bumps liquidjs-lib and ldk to latest (tag 6.0.1).
* add a `descriptors` package letting to evaluate output script descriptors templates.
* templates are tested in `tests/descriptors.spec.ts` file.

Marina descriptors and API documentation: [descriptors/doc.md](https://github.com/louisinger/marina/blob/template-parser/src/descriptors/doc.md).

it closes #320

@tiero please review


src/descriptors/doc.md
```diff
@@ -0,0 +1,82 @@
...
+const redeemScript = result.redeemScript // OP_O .... 
+```
+
+> The `result` object value depends on the Context! 
+
+## API
+
+All descriptors can accept `xpub...` token, the public keys are replaced before parsing the template.
+
+### Raw
```
tiero, line 35: `raw` should only be used at top level imho, like original descriptors and...shouldn't be the scriptPubKey? (not redeem script as you mention)

tiero, line 35: `raw` should only be used at top level imho, like original descriptors and...shouldn't be the scriptPubKey? (not redeem script as you mention)

src/descriptors/doc.md
```diff
@@ -0,0 +1,82 @@
...
+Returns:
+
+* `raw(0123456789)` redeem script as `witnesses`.
+* the segwit v0 script `(OP_0 + hash(raw(0123456789)))` as `redeemScript`.
+
+### Eltr
+
+It lets to create [P2TR](https://en.bitcoin.it/wiki/Taproot) payments. The descripor expects 2 inputs:
+
+1. The internal taproot public key. This can be replaced by an `xpub` token in order to increment it according to actual derivation path. Will fail if this is not a 64bits hex string.
```
tiero, line 72: we should maybe add `H_POINT` keyword in preprocessing, so we map to unspendable internal pub key for script-path-only P2TR for the three cases

tiero, line 72: we should maybe add `H_POINT` keyword in preprocessing, so we map to unspendable internal pub key for script-path-only P2TR for the three cases

louisinger:
changes included in #345

PR #335, opened 2022-04-18
Makes popup prompt appear on top right of the screen

bordalix:
Since `browser.windows.create` options doesn't allow `right`, we need
to calculate `left` with window width, popup width, and some margin.

@tiero please review


tiero:
Seems to fragile. Would do like metamask here

https://github.com/MetaMask/metamask-extension/blob/3732c5f71ef60d6913f81e4d4623b9ddee11e3c0/app/scripts/lib/notification-manager.js#L36-L68

bordalix:
b387392 should fix it.

@tiero please review

PR #345, opened 2022-04-29
Marina CustomScript account

### LINKED ISSUE #318: Support template based account
related to: https://github.com/vulpemventures/marina/issues#250

Update marina's broker and provider according to: https://github.com/vulpemventures/marina-provider/issues/18

### LINKED ISSUE #320: Implement script type `Template`

### LINKED ISSUE #321: implement a Custom "Template" IdentityInterface
related to #250 

A `TemplateIdentityInterface` aims to support all script type described by a `Template` (a scripts high level language like miniscript). 

- [ ] implement LDK `IdentityInterface`
- [ ] Script type definition via Template #320
- [ ] Bip32 custom deterministic path #319

### LINKED ISSUE #322: Sign taproot input
related to #250 

with the introduction of Template based accounts, Marina should be able to sign a taproot input.

**Key path**
sign with schnorr, this is pretty similar of witness v0.

**Script path**
script path defines: spend a taproot coin via the leaf **L**

two cases:

1. either Marina has all the information needed to create the witness stack from **script(L)**: it means that **script(L)** only requires marina privkeys.
2. or Marina need an extra array of data **I**.

(1) the web dev could for instance specify the leaf (mapped with input index) in `signTransaction`.
(2) in that case, maybe better to let the user sign the pset and build the custom **I**. Note that we could add another `marina.getLeafInfo(L)` returning the path, the script and the taproot internal key (all the data needed to fill a script path witness stack).

louisinger:
**This PR implements the new provider described here: https://github.com/vulpemventures/marina-provider/pull/20**

It lets marina to send and receive funds on descriptors-based accounts.

**Custom script account**

The main change is the new CovenantAccount, implementing the `Account` interface via a new `IdentityInterface`. The `CovenantIdentity` is using a descriptors-like language to describe the account script type. 

utxos and transactions updating is already handled by the current version of updater. The major change about this is the new `selectedAccount` member in the marina provider: it lets to implement new `marina.useAccount`, then updater and restorer in background work like the MainAccount ones. In Popup UI, "click on update" will launch an update for all the accounts.

Accounts are identified by their namespace or `AccountID`. The main account ID is a constant: "main". The covenant account ID aims to derive the wallet master key (using [slip13](https://github.com/satoshilabs/slips/blob/master/slip-0013.md)). Thus, with mnemonic + namespace + template(s), the user is able to restore its covenants accounts.

**User interface changes**

This PR does not modify the UX but some changes are introduced by "multiple accounts paradigm": 
- new settings/accounts page -> let u know what's the accounts set up on your extension.
- on main page, the balances values = sum of all accounts balances.
- on "send" page, balances values = sum of all **auto-spendable** accounts balances. auto-spendable means that marina can easily compute the sig needed to sign one of the account's coin. see `script-analyzer.ts`.
- no changes in receive, only main account addresses can be generated in popup.
- no changes in onboarding flow, to restore a covenant account u have to recreate it.

**How to test**

- build, install and onboard your mnemonic.
- visit vulpem.com or any other websites allowing you to run the marina provider.
- run `marina.enable()` if its the first time you are using the provider there.
- run `marina.createAccount('test')` -> will derive your mnemonic according to slip13. **test is the new account ID**.
- run `marina.useAccount('test')` -> switch from "main" to "test".
- run `marina.importTemplate({ type: "marina-descriptors", template: "eltr($test, { asm($test OP_CHECKSIG), asm(OP_FALSE) })" })
`
- run `marina.getNextAddress` -> u have a taproot address, try to faucet it. the template contains $test OP_CHECKSIG which is an auto-spendable leaf so u can spend this coin in Popup (or using the provider).

**ecclib.ts trick to bump LDK**

TL;DR
- **ecclib.ts** shouldn't be imported in other src/*.ts files. 
- **tiny-secp256k1** will be replaced by the export of `ecclib.ts` via webpack.
- no changes for developer, he can use tiny-secp256k1 like a normal dependencie (but ecclib will be used under the hood).

In order to support BIP341 (or segwitv1) changes, we need to bump the latest LDK lib. This new version requires the injection of `tiny-secp256k1` lib in some packages. The `import * as ecc from 'tiny-secp256k1';` is doing that job. 

However, the secp256k1 implementation works with wasm and marina extension is building three JS contexts: `popup`, `background` and `content`. for security reasons it is not possible to inject .wasm files directly in content scripts. That's why we need to import base64 wasm (.wasm files encoded as a base64 string which can be embedded in the javascript files).

To replace `import 'tiny-secp256k1'` by the "base64 wasm string" we are using the `src/ecclib.ts` file + webpack aliases. This is totally transparent for developers (import 'tiny-secp256k1` like the other dependencies).

it closes #320 
it closes #322 
it closes #321 
it closes #318 
it closes #325 


src/application/redux/containers/address-amount.container.ts
```diff
@@ -1,18 +1,19 @@
...
 import { selectBalances } from '../selectors/balance.selector';
-import { selectMainAccount } from '../selectors/wallet.selector';
+import {
+  selectAllAccountsIDsSpendableViaUI,
+  selectMainAccount,
+} from '../selectors/wallet.selector';
 
 const mapStateToProps = (state: RootReducerState): AddressAmountProps => ({
-  account: selectMainAccount(state),
+  changeAccount: selectMainAccount(state),
```
tiero, line 13: not clear why now we calling it `changeAccount` (need to finish my review yet) 

How do we know the current "active" account?

louisinger: When u send coin from popup, after selection the asset to send. Marina will select utxos from all accounts. So if u own LBTC on a covenant account & on the main account, u are not able to "select" your utxos (or filter utxos by account(s)). That's why if change is needed, the main account is used.

In my mind this is a temporary solution. I think the user should be able to select the "change account" during "select asset fees" step.

I've renamed the variable because that account is the one used to generate change addresses (including taxi one).

louisinger:
> Independently of the network chosen, the extension is always trying to access http://localhost:8000/assets (taxi regtest url).

Right. I was thinking we do that already on master, will remove the regtest update on non-regtest networks

bordalix:
> > Independently of the network chosen, the extension is always trying to access http://localhost:8000/assets (taxi regtest url).
> 
> Right. I was thinking we do that already on master, will remove the regtest update on non-regtest networks

Just to be clear, we do pool periodically (1 minute) taxi url. But we do it to the corresponding taxi url for the chosen network. The issue here is that it is ignoring the chosen network and always pooling the regtest url.

src/application/redux/sagas/main.ts
```diff
@@ -9,15 +10,17 @@ import {
...
 import { updateAfterEachLoginAction, watchUpdateTask } from './updater';
 import { watchStartDeepRestorer } from './deep-restorer';
-import { NetworkString } from 'ldk';
+import type { NetworkString } from 'ldk';
 import { selectTaxiAssetsForNetwork } from '../selectors/taxi.selector';
 
 function* fetchAndSetTaxiAssets(): SagaGenerator<void, string[]> {
   yield* fetchTaxiAssetsForNetwork('liquid');
   yield* fetchTaxiAssetsForNetwork('testnet');
+  yield* fetchTaxiAssetsForNetwork('regtest');
```
bordalix, line 23: Remove?

src/presentation/wallet/send/end-of-flow.tsx
```diff
@@ -91,13 +91,18 @@ const EndOfFlow: React.FC<EndOfFlowProps> = ({
       // add unconfirmed utxos from change addresses to utxo set
       if (unconfirmedOutputs && unconfirmedOutputs.length > 0) {
         await dispatch(
-          await addUnconfirmedUtxos(tx, unconfirmedOutputs, accounts[0].getAccountID(), network)
+          await addUnconfirmedUtxos(
+            tx,
+            unconfirmedOutputs,
+            signerAccounts[0].getAccountID(),
```
bordalix, line 97: This is crediting unconfirmed utxos to the main account, not to the signing account. Is this correct?

If so, can we change it to `MainAccountID` exported from `src/domain/account.ts`?

src/domain/script-analyser.ts
```diff
@@ -0,0 +1,247 @@
+// analyse a script to know what it need as input
```
tiero, line 1: This file looks like someting LDK can export eventually

src/domain/covenant-identity.ts
```diff
@@ -0,0 +1,429 @@
...
+import * as ecc from 'tiny-secp256k1';
+import type { BIP32Interface } from 'bip32';
+import BIP32Factory from 'bip32';
+import { mnemonicToSeedSync } from 'bip39';
+import type { ScriptInputsNeeds } from './script-analyser';
+import { analyzeTapscriptTree } from './script-analyser';
+
+// slip13: https://github.com/satoshilabs/slips/blob/master/slip-0013.md#hd-structure
+function namespaceToDerivationPath(namespace: string): string {
+  const hash = crypto.sha256(Buffer.from(namespace));
```
tiero, line 42: We should also fingerprint the template itself imho 

Sort of versioning of the contract/template changes overtime

src/domain/covenant-identity.ts
```diff
@@ -0,0 +1,429 @@
...
+      args.chain
+    );
+
+    const masterBlindingKeyNode = SLIP77Factory(args.ecclib).fromSeed(seed);
+
+    super({
+      ...args,
+      opts: {
+        namespace: args.opts.namespace,
+        template: args.opts.template,
```
tiero, line 254: what about changeTemplate

tiero:
Tried restore an old menmonic, I got 

image

src/domain/covenant-identity.ts
```diff
@@ -0,0 +1,429 @@
...
+        ...covenantDescriptors,
+        masterPublicKey,
+        masterBlindingKey,
+      },
+    })
+  )(restorerOpts);
+}
+
+function makePath(isChange: boolean, index: number): string {
+  return `${isChange ? 1 : 0}/${index}`;
```
tiero, line 428: why we not return the full bip32 path? m/A'/B' etc..

src/domain/covenant-identity.ts
```diff
@@ -0,0 +1,429 @@
...
+  getAddress(isChange: boolean, index: number): TaprootAddressInterface {
+    const template = this.getTemplate(isChange);
+    const descriptorResult = evaluate(this.getContext(isChange, index), template);
+    const outputScript = descriptorResult.scriptPubKey().toString('hex');
+    if (!descriptorResult.taprootHashTree) throw new Error('taprootHashTree is missing');
+    return {
+      ...this.outputScriptToAddressInterface(outputScript),
+      result: descriptorResult,
+      derivationPath: makePath(isChange, index),
+      publicKey: this.deriveMasterXPub(isChange, index, true),
```
tiero, line 157: I guess this should be empty for this type of addresses

We should eventually create a dedicated field for the taproot address and we will call `interalPubKey` to be explicit on the meaning of it.

src/domain/covenant-identity.ts
```diff
@@ -0,0 +1,429 @@
...
+  return `m/${A}/${B}/${C}/${D}`;
+}
+
+function makeBaseNodeFromNamespace(m: BIP32Interface, namespace: string): BIP32Interface {
+  const path = namespaceToDerivationPath(namespace);
+  return m.derivePath(path);
+}
+
+export interface TaprootAddressInterface extends AddressInterface {
+  result: Result;
```
tiero, line 57: Not sure is meaningful returning this data structure to user/dev

src/domain/covenant-identity.ts
```diff
@@ -0,0 +1,429 @@
...
+  tapscriptNeeds: Record<string, ScriptInputsNeeds>; // scripthex -> needs
```
tiero, line 58: Depends on the tooling we will offer to users (via LiquidJS p2tr or something on top with LDK) but I feel we need to tell user also the "order" of leafs.

Ie. how to know which script was the "first" in the array I inserted? Clients would need to compile out of band the inserted TREE in the template to know the script hex

src/domain/covenant-identity.ts
```diff
@@ -0,0 +1,429 @@
...
+}
+
+export interface CovenantDescriptors {
+  namespace: string;
+  template?: string;
+  changeTemplate?: string;
+  isSpendableViaUI?: boolean;
```
tiero, line 65:
```suggestion
  isSpendable?: boolean;
```

src/domain/covenant-identity.ts
```diff
@@ -0,0 +1,429 @@
+import {
```
tiero, line 1: How bad would be renaming it "CustomScriptIdentity" ? :)

tiero, line 1: How bad would be renaming it "CustomScriptIdentity" ? :)

src/domain/covenant-identity.ts
louisinger: isSpendable may mislead future devs IMO, what do u think of `isSpendableByMarina` ?

src/domain/covenant-identity.ts
louisinger: > Depends on the tooling we will offer to users (via LiquidJS p2tr or something on top with LDK) but I feel we need to tell user also the "order" of leafs.

We agreed on do not expose `tapscriptNeeds` member to dev/user isn't it ?
PS: I see it as a cached data for signing step (the Record data structure has been choosen to work later with all types of descriptors). The idea is to do no re-compute two times the "needs" for a given script.

> Ie. how to know which script was the "first" in the array I inserted? Clients would need to compile out of band the inserted TREE in the template to know the script hex

Right. That's why I prefer the binary tree structure. `liquidjs.HashTree` induces the order of the leaves (moreover that's also a binary tree used in `eltr` templates)

tiero: Yep, discard this comment, came before of our review call

src/domain/covenant-identity.ts
tiero: Yep better

src/domain/account.ts
```diff
@@ -75,6 +110,61 @@ export function createMnemonicAccount(data: MnemonicAccountData): MnemonicAccoun
...
+}
+
+const factories = new Map<
+  AccountType,
+  (encryptedMnemonic: EncryptedMnemonic, data: any) => Account
+>()
+  .set(AccountType.MainAccount, createMainAccount)
+  .set(AccountType.CustomScriptAccount, createCustomScriptAccount);
+
+export const createAccount = (encryptedMnemonic: EncryptedMnemonic, data: AccountData): Account => {
```
tiero, line 160: would rather call something more explicit? ie. `accountFromMnemonicAndData`

package.json
```diff
@@ -35,7 +35,7 @@
     "ldk": "git+https://github.com/louisinger/ldk.git#descriptors-and-script-analyzer",
     "lodash.debounce": "^4.0.8",
     "lottie-web": "^5.7.8",
-    "marina-provider": "^1.4.10",
+    "marina-provider": "1.4.10",
```
tiero, line 38: should be 1.5 ?

tiero:
WRT which account to use for change when cross-spending utxos accross accounts?

Best solution so far is to use the account chosen with `useAccount` API, below the equivalent UI component 

Home empty - drop down menu

tiero:
Other feedback on UI: best to "hide" it in settings, switching address type/account is not really something user want to do often.

Also we should show the current used account in the receive tab and in send flow ("we using this $account for the change address)

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.4.5",
   "private": true,
   "scripts": {
-    "start": "NETWORK=regtest webpack --watch --config webpack.dev.js",
+    "start": "NETWORK=regtest webpack --config webpack.dev.js",
```
tiero, line 6: Why remove watch?

louisinger: Better to put it in the config file I think (and it lets us to set up the "watchOptions" field also)

tiero:
https://user-images.githubusercontent.com/3596602/169332441-326f4624-7363-4ab0-a5f4-20480f91c4fb.mov

tiero:
Is Accounts tab in UI setting binded to actual account selector? I switched from `test` to `main`, and `await marina.getSelectedAccount()` is still returning me` test`

tiero:
I get invalid template on `marina.importTemplate("eltr($test, { asm($test OP_CHECKSIG), asm(OP_FALSE) })")`

louisinger:
> Is Accounts tab in UI setting binded to actual account selector? I switched from `test` to `main`, and `await marina.getSelectedAccount()` is still returning me` test`

louisinger:
> Is Accounts tab in UI setting binded to actual account selector? I switched from `test` to `main`, and `await marina.getSelectedAccount()` is still returning me` test`

the settings/account selector let you choose the changeAccount not the "selected one". selectedAccount is a "local" variable in the webapp context, so each webapp has a different selectedAccount

src/application/utils/transaction.ts
```diff
@@ -119,16 +152,39 @@ export async function blindAndSignPset(
...
+    }
+  }
 
-  const decodedPset = decodePset(signedPset);
-  if (!decodedPset.validateSignaturesOfAllInputs()) {
-    throw new Error('PSET is not fully signed');
+  for (let i = 0; i < pset.txInputs.length; i++) {
+    const input = pset.data.inputs[i];
+    // we need to use special finalizer in case of tapscript
+    if (atLeastOne(input.tapLeafScript) && atLeastOne(input.tapScriptSig)) {
```
tiero, line 167: It can happen a tapscript branch does not have any signature tho: ie. a dumb `OP_ADD 3 OP_EQUAL` does not need any signature there, but needs two elements on the stack

src/application/utils/transaction.ts
```diff
@@ -119,16 +152,39 @@ export async function blindAndSignPset(
...
+      pset.finalizeInput(i, (_, input) => {
+        return {
+          finalScriptSig: undefined,
+          finalScriptWitness: witnessStackToScriptWitness([
+            ...input.tapScriptSig!.map((s) => s.signature),
+            input.tapLeafScript![0].script,
```
tiero, line 173: how do we ensure the first element is always the correct one? ie. its always spendable?

src/application/utils/transaction.ts
louisinger: this "special finalizer" is mainly here for the "auto-spendable by marina" case. If the leaf does not need signature, it can't be finalized by marina (user needs to use a special finalizer)

src/application/utils/transaction.ts
louisinger: I think we can assume that, at finalize step, signature has been verified. The idea of the finalizer is to use it as "default" to handle the simple cases: simple leaf spendable by marina (basically the $key CHECKSIG script only). This is only used in UI SEND & `marina.sendTransaction` preventing user to coin-select a "complex" leaf.

tiero:
There is some problems with  incrementing the index of the derivation path: if you try call the getNextAddress multiple time will return always the index 1 and 2

louisinger:
related PR: https://github.com/louisinger/marina/pull/1

package.json
```diff
@@ -24,18 +24,19 @@
     "@tailwindcss/postcss7-compat": "^2.1.2",
     "autoprefixer": "^9.8.6",
     "axios": "^0.21.2",
+    "bip32": "^3.0.1",
     "bip39": "^3.0.4",
     "bitcoinjs-message": "^2.2.0",
```
tiero, line 29: add bip174 from your fork

tiero:
on MainAccount f I do `await marina.getAccountInfo()` will get `Uncaught Error: Account undefined not found`

PR #346, opened 2022-05-03
Marina Web Provider: Broadcast transaction

### LINKED ISSUE #341: Add `marina.broadcastTransaction`
User can find this API useful to let Marina handle pushing tx to the network and eventually credit virtual change output if recognized

bordalix:
Closes #341 

When broadcasting a transaction:
- inspects inputs to find own coins and lock them;
- inspects outputs to find change addresses and credit it them.

@tiero please review

### DIFF
--- a/src/application/utils/utxos.ts
+++ b/src/application/utils/utxos.ts
@@ -8,14 +8,8 @@ import {
 } from 'ldk';
 import { address, networks, Transaction } from 'liquidjs-lib';
 import { RawHex } from 'marina-provider';
-import { RootReducerState } from '../../domain/common';
+import { Account } from '../../domain/account';
 import { UnconfirmedOutput } from '../../domain/unconfirmed';
-import { selectNetwork } from '../redux/selectors/app.selector';
-import {
-  selectAllAccounts,
-  selectAllAccountsIDs,
-  selectUtxos,
-} from '../redux/selectors/wallet.selector';
 
 export const toStringOutpoint = (outpoint: { txid: string; vout: number }): string => {
   return `${outpoint.txid}:${outpoint.vout}`;
@@ -43,8 +37,10 @@ export interface UtxosFromTx {
 
 // given a signed tx hex, get selected utxos and change utxos
 export const getUtxosFromTx = async (
+  accounts: Account[],
+  coins: UnblindedOutput[],
+  network: NetworkString,
   signedTxHex: RawHex,
-  state: RootReducerState
 ): Promise<UtxosFromTx> => {
   const tx = Transaction.fromHex(signedTxHex);
   const txid = tx.getId();
@@ -52,9 +48,6 @@ export const getUtxosFromTx = async (
   // get selected utxos used in this transaction:
   // - get all coins from marina
   // - iterate over all transaction inputs and check if is a coin of ours
-
-  // get all coins from marina
-  const coins = selectUtxos(...selectAllAccountsIDs(state))(state);
 
   // array to store all utxos selected in this transaction
   const selectedUtxos: UnblindedOutput[] = [];
@@ -80,8 +73,8 @@ export const getUtxosFromTx = async (
 
   // get all marina addresses (from all accounts)
   let allAddresses: AddressInterface[] = [];
-  for (const account of selectAllAccounts(state)) {
-    const identity = await account.getWatchIdentity(selectNetwork(state));
+  for (const account of accounts) {
+    const identity = await account.getWatchIdentity(network);
     allAddresses = [...allAddresses, ...(await identity.getAddresses())];
   }
 
@@ -91,7 +84,7 @@ export const getUtxosFromTx = async (
       // get address from output script
       const addressFromOutputScript = address.fromOutputScript(
         Buffer.from(out.script),
-        networks[selectNetwork(state)]
+        networks[network]
       );
       // iterate over marina addresses
       for (const addr of allAddresses) {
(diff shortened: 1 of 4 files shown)


tiero:
Should we need to add to marina-provider interface as well?

tiero:
Can you make a build and attach here? so we ask other to test it out before merge

bordalix:
> Should we need to add to marina-provider interface as well?

PR https://github.com/vulpemventures/marina-provider/pull/23

src/content/marina/marinaBroker.ts
```diff
@@ -326,6 +328,102 @@ export default class MarinaBroker extends Broker {
...
+          // will need tx to lock selected utxos and credit unconfirmed utxos
+          const tx = Transaction.fromHex(signedTxHex);
+
+          // lock selected utxos used in this transaction:
+          // - get all coins from marina
+          // - iterate over all transaction inputs and check if is a coin of ours
+          // - if any, lock selected utxos
+
+          // get all coins from marina
+          const coins = selectUtxos(MainAccountID)(state);
```
tiero, line 349: Do we need a selector to retrieve all possible accounts here cc/ @louisinger

bordalix: What about marina.getCoins?

louisinger: You can use `selectUtxos(selectAllAcountsIDs(state)...)(state)`

src/content/marina/marinaBroker.ts
```diff
@@ -326,6 +328,102 @@ export default class MarinaBroker extends Broker {
...
+
+          // array to store all utxos selected in this transaction
+          const selectedUtxos: UnblindedOutput[] = [];
+
+          // find all inputs that are a coin of ours
+          tx.ins.forEach((_in) => {
+            const { hash, index } = _in; // txid and vout
+            const txid = hash.reverse().toString('hex');
```
louisinger, line 357: nit:To avoid weird errors later, I'm used to combine `Array.slice()` with reverse func. `hash.slice(). reverse()` ensures that hash variable is not mutated.

src/content/marina/marinaBroker.ts
```diff
@@ -326,6 +328,102 @@ export default class MarinaBroker extends Broker {
...
+            }
+          }
+
+          // credit change utxos:
+          // - get all addresses used by marina
+          // - iterate over all transaction outputs and check if belongs to marina
+          // - if any, add unconfirmed utxos to utxo set
+
+          // get all marina addresses (reversed for performance reasons)
+          const xpub = await selectMainAccount(state).getWatchIdentity(network);
```
louisinger, line 379: It's ok now but ideally it should work with multiple accounts "paradigm". Maybe u could select all accounts' here ?

louisinger (commented):
do you think it is possible to isolate the logic you have implemented in broadcastTransaction and use it also in the popup send flow ? we also broadcast a tx there at the end of the user journey (the idea is to remove the "uncomfirmedUtxos" logic in end-of-flow if possible)

src/content/marina/marinaBroker.ts
bordalix: Fixed by 33dafc3

src/content/marina/marinaBroker.ts
bordalix: Fixed by 8d2c329

src/content/marina/marinaBroker.ts
```diff
@@ -326,6 +328,102 @@ export default class MarinaBroker extends Broker {
...
```
tiero, line 379: It's more appropriate calling this `identity` rather than xpub

src/content/marina/marinaBroker.ts
bordalix: Fixed by 7244ddb

src/content/marina/marinaBroker.ts
bordalix: Fixed by 7244ddb

bordalix:
> do you think it is possible to isolate the logic you have implemented in broadcastTransaction and use it in the popup send flow ? we also broadcast a tx there at the end of the user journey (the idea is to remove the "uncomfirmedUtxos" logic in end-of-flow if possible)

Not 100% dry code, but better with 41fc7da

bordalix:
Build for testing:

[marina-0.4.5-RC2.zip](https://github.com/vulpemventures/marina/files/8630618/marina-0.4.5-RC2.zip)

src/application/utils/utxos.ts
```diff
@@ -1,22 +1,149 @@
...
+import { address, networks, Transaction } from 'liquidjs-lib';
+import { RawHex } from 'marina-provider';
+import { RootReducerState } from '../../domain/common';
 import { UnconfirmedOutput } from '../../domain/unconfirmed';
+import { selectNetwork } from '../redux/selectors/app.selector';
+import {
+  selectAllAccounts,
+  selectAllAccountsIDs,
+  selectUtxos,
+} from '../redux/selectors/wallet.selector';
```
tiero, line 19: not entirely convinced we should import redux methods in an util method, just for separate concerns and re-utilize the utils in other context? What if passing clousure as parameters so we can pass it from the top

toughts @louisinger?

louisinger: +1, better to keep selectors outside of functions (mainly to be able to launch them in the `container` files for UI)

bordalix: Fixed by c900ca6

tiero (approved):

PR #347, opened 2022-05-05
Use icons from mempool

### LINKED ISSUE #334: Retrieve asset icons from Blockstream registry
https://assets.blockstream.info/icons

bordalix:
Closes #334

This PR:

- Updates the pull-asset-registry.js script:
  - fetches all asset hashes supported by Mempool from [github](https://raw.githubusercontent.com/Blockstream/asset_registry_db/master/icons.json) 

- Updates constants.ts:
  - every supported asset will have its icon served from https://liquid.network/api/v1/asset/${assetHash}/icon

Warning: this creates a dependency from https://liquid.network, who, until today, didn't fix the problem with https://liquid.network/api/v1/assets/icons

@tiero please review


scripts/pull-asset-registry.js
```diff
@@ -16,14 +16,10 @@ async function main() {
     writeToFile(lightNitePath, lightNiteAssetHashes);
 
     // Blockstream
-    const blockstreamAssetHashes = Object.values(res.data)
-      .filter(asset => asset.entity.domain.includes("blockstream.com"))
-      .map(asset => asset.asset_id);
-
+    res = await axios.get('https://raw.githubusercontent.com/Blockstream/asset_registry_db/master/icons.json');
+    const blockstreamAssetHashes = Object.keys(res.data);
```
tiero, line 20: mmm blockstreamAssetHashes means they are issued by blockstream hence before I was checking in the domain field of the asset if is from them

bordalix: Yes, with the only purpose of having a default icon ('blockstream.png') for each asset where we didn't have the original icon. But, previously this would find 17 assets which now are ALL present in the new [json endpoint](https://raw.githubusercontent.com/Blockstream/asset_registry_db/master/icons.json), so we don't need this default icon logic anymore.

So now this variable means "icons we have from mempool because they support all blockstream asset hashes present in the new json endpoint".

tiero: Yes so the variable name should change imho, but maybe what we should do is dropping this script entirely? I mean if we can now dinamically fetch an icon via network call, no need to bundle anything imho

bordalix: We need to know which asset hashes are available on https://liquid.network/api/v1/asset in order to implement the logic of showing an UNKNOW_IMG in the case they are not (see [getAssetImage()](https://github.com/bordalix/marina/blob/use_icons_from_mempool_334/src/application/utils/constants.ts#L65))

tiero: What if we show Unknown Image anytime we either a) Get a 404 not found error b) the API is down?

I guess I would only retain hardcoded fallback for the real featured one (L-BTC & USDT) that's it, to lower the bundle as much as possible?

bordalix: Hummmm, that would mean getAssetImage() would have to make a fetch for the image, wait for the return value, and then return the image path based on the return code from the fetch. But then the browser would make the fetch also, which would mean double hits for every image. Not very elegant.

What we can do, though, is to use something of this kind on every image tag:

```
<img src="https://remote.path/image.png" onerror="this.src='images/unknown.png'">
```

What do you think?

tiero: I like it! let's do it (it removes code and complexity and rely on native browser apis)

I would still pass a function to `src` where it will check if the asset hash is the featured ones (LBTC, USDT) and save a network call?

bordalix: Fixed by dacac83

Assets size reduction:
  index.js (2.68 MiB) => 2.66 MiB
  background-script.js (1.34 MiB) => 1.32 MiB
  content-script.js (1.35 MiB) => 1.33 MiB

src/application/utils/constants.ts
```diff
@@ -12,60 +10,44 @@ export const feeLevelToSatsPerByte: { [key: string]: number } = {
...
-const lbtcIcon = makeRemoteImagePath(networks.liquid.assetHash);
-assetHashToImagePath.set(networks.testnet.assetHash, lbtcIcon);
-assetHashToImagePath.set(networks.regtest.assetHash, lbtcIcon);
-
-// set icons for usdt in testnet
-const usdtAssetHash = {
-  liquid: 'ce091c998b83c78bb71a632313ba3760f1763d9cfcffae02258ffa9865a37bd2',
-  testnet: 'f3d1ec678811398cd2ae277cbe3849c6f6dbd72c74bc542f7c4b11ff0e820958',
+// featured assets
+const featuredAssets = {
```
tiero, line 14: what about BTC

src/presentation/wallet/home/index.tsx
```diff
@@ -91,7 +91,7 @@ const HomeView: React.FC<HomeProps> = ({
           <Balance
             assetHash={lbtcAssetHash}
             assetBalance={fromSatoshiStr(assetsBalance[lbtcAssetHash] ?? 0)}
-            assetImgPath="assets/images/liquid-assets/liquid-btc.svg"
+            assetImgPath={getAssetImage(lbtcAssetHash)}
```
tiero, line 94: I think here was fine to hardcode

bordalix: I removed the entire `assets/images/liquid-assets` folder.
By using `lbtcAssetHash` is locked to L-BTC.

src/application/utils/constants.ts
```diff
@@ -12,60 +10,44 @@ export const feeLevelToSatsPerByte: { [key: string]: number } = {
...
+// translates testnet and regtest asset hashes to mainnet asset hash
+const featuredAssetsMap = new Map<string, string>();
+featuredAssetsMap.set(networks.testnet.assetHash, networks.liquid.assetHash);
+featuredAssetsMap.set(networks.regtest.assetHash, networks.liquid.assetHash);
+featuredAssetsMap.set(featuredAssets.usdt.testnet, featuredAssets.usdt.liquid);
+featuredAssetsMap.set(featuredAssets.lcad.testnet, featuredAssets.lcad.liquid);
+
+// given an asset hash, return url for image path
+const getImagePath = (hash: string): string => {
+  const remoteServer = 'https://liquid.network/api/v1/asset';
```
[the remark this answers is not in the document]
bordalix: Yes, it could make sense: after all they are used all the time.
Ok, I'm going to bring back the icons folder and change that.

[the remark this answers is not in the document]
tiero: Yes! The dynamic loading helps for minor tokens and is future proof so we do no need to manually bundle them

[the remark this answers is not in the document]
bordalix: Fixed by 414ecfd

PR #349, opened 2022-05-10
lightning network

tiero:
## Overview

This propose a change to the UI/UX as specified in #255  and adds Lightning Network payments via Boltz submarine swaps.
NOTICE: this is just a PoC, need serious work and cleanup to reach production.

## How to test

Assuming you restore/create a wallet, pick password `ciaociao` you should be able to

0. Select Testnet
1. Click on Receive
2. Select Liquid Bitcoin
3. Select Lightning Network
4. Enter an amount > 0.0005
5. Pay an invoice with HTLC.me in
6. After invoice paid, Marina will broadcast Liquid tx and after a minute you will have the new LBTC deposited 
## TODOs

- [x] Show Modal to pick Liquid or Lightning when sending/receiving L-BTC
- [ ] Fix the the Modal be a Bottom Sheet Modal as per design
- [ ] Reverse Submarine flow: Use Redux to persist state and make of use of two separated pages 
  - [ ] Enter Amount page
  - [ ] Show Invoice/Success/Failure page
  - [ ] Prompt user password to Unlock and Sign the Boltz claim transaction
- [ ] Re-implement `constructClaimTransaction` with Psbt class in order to exploit LDK's identity (now it extracts keys, drop this) 
- [ ] Make correct transaction size estimation based on the HTLC swap script


tiero:
For the Submarine flow instead, I guess the new UX should be:

1. Click on Send
2. Select Liquid Bitcoin
3. Select Lightning Network
4. Enter an amount user want to withdraw > 0.0005
6. Paste the LN invoice inside the next screen in marina
7. Send the request to boltx which will answer with the lockup address
8. Send L-BTC funds to that
 
After funds are locked onchain, Boltz will pay the invoice on our behalf

bordalix:
> 4. Enter an amount user want to withdraw > 0.0005

I believe there's also a max amount of 0.1 BTC on Boltz.

Do you confirm?

tiero:
> > 4. Enter an amount user want to withdraw > 0.0005
> 
> I believe there's also a max amount of 0.1 BTC on Boltz.
> 
> Do you confirm?

Correct! We should shot those limits in the UI eventually/show errors

PR #350, opened 2022-05-12
Add dispatchEvent for #installed and #updated

### LINKED ISSUE #337: onInstalled event listener
We need an event listener to catch user's new marina installation.
Like https://stackoverflow.com/questions/52347230/detect-when-chrome-extension-was-installed-without-inline-installation/52353104#52353104

tiero:
This closes #337 

Please @bordalix @huseyinsaritas review

### DIFF
--- a/src/background/background-script.ts
+++ b/src/background/background-script.ts
@@ -13,6 +13,7 @@ import {
   OpenPopupMessage,
   PopupName,
 } from '../domain/message';
+import Marina from '../inject/marina/provider';
 import { POPUP_RESPONSE } from '../presentation/connect/popupBroker';
 import { INITIALIZE_WELCOME_ROUTE } from '../presentation/routes/constants';
 import { periodicTaxiUpdater, periodicUpdater } from './alarms';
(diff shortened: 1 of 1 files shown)


tiero:
[marina-0.4.5.zip](https://github.com/vulpemventures/marina/files/8679367/marina-0.4.5.zip)

tiero:
Good point, sorry to not have tested that.

What we can do then is maybe have the bg script to `window.postMessage`? then the content script can have a listener there for such message and dispatch the event? (like we do with the broker)

bordalix:
Inject script already emits an event when injected (`marina#initialized`).

I'm doing some experiments on what can we achieve with it.

bordalix:
I think this is simply not possible.

For the extension to be able to inform the web app it was installed or updated, the content script must be injected on the web app's DOM. But, accordingly to the manifest documentation [1], content scripts are only injected when the page (`Document.readyState`) emits one of these 3 events:
- loading
- interactive
- completed

So, if a web app is fully loaded, and we install or update the extension, the content script will not be injected to the DOM, since none of these 3 events is emitted.

[1] See `run_at` here: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts

PR #351, opened 2022-05-12
Fix address reuse if recipient asset !== L-BTC

### LINKED ISSUE #340: Change BTC not tracked by wallet
Take this transaction
https://blockstream.info/liquidtestnet/tx/af74a83c2374ce906bd7e5b09937a3866cafd9fc2ba3df005422b0df052289db#blinded=

For Marina this is a USDT outbound transaction, but LBTC balance is zero instead of `146.31269123` which seems to be the L-BTC change

image

- Multi-asset transaction makes address reuse between change asset output & btc change output
- L-BTC change output is unconfidential
- From Marina can't generate unblinding URL somehow

louisinger:
related to https://github.com/vulpemventures/ldk/pull/119

The LDK fix lets us reuse the `changeAddressGetter` used in taxi pset. If the recipient asset is not L-BTC, a specific change address will be generated.

This PR also includes a small fix in txUpdater saga: now, we force re-updating of transactions where `transfers.length === 0`. It lets to fix bad Inbound/Outbound displaying in case of wallet with gap.

it closes #340 

@tiero @Janaka-Steph please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "decimal.js": "^10.2.1",
     "formik": "^2.2.6",
     "google-protobuf": "^3.15.8",
-    "ldk": "^0.4.12",
+    "ldk": "0.4.12",
     "lodash.debounce": "^4.0.8",
     "lottie-web": "^5.7.8",
     "marina-provider": "^1.4.9",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "decimal.js": "^10.2.1",
     "formik": "^2.2.6",
     "google-protobuf": "^3.15.8",
-    "ldk": "git+https://github.com/louisinger/ldk.git#hotfix-crafsinglerecipientpset",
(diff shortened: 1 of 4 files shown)


tiero:
Shall we use ldk@next in marina? since we are going to support nicely liquidJS@6.x.x?

louisinger:
> Shall we use ldk@next in marina? since we are going to support nicely liquidJS v6.x.x?

I would wait #345 (to test is everything is working well with update)

tiero:
Test are failing?

tiero:
> in case of wallet with gap.

Not sure problem happens for a gap limit reached TBH, I guess would happen anyway

tiero:
having problem trying to restore seed with a build for production

```
Restoration failed
Unsuported witness version (undefined), only 0 (blech32) or 1 (blech32m) are supported
```
image

louisinger:
included in #345

PR #354, opened 2022-06-01
AssetIcon component 

louisinger:
This PR cherry-pick [AssetIcon commit](https://github.com/vulpemventures/marina/pull/345/commits/9b8d8b4c732850b49687840252b379981b1198ff) from #345 

- use `AssetIcon` to load asset's icon and do not re-assign img source if error occurs (launching a loop causing image blinking).
- use `React.memo` to do not re-mount asset component each time we render the page (and so do no re-send img requests). Also use React.memo in any others pure functional components (TxIcon, Balance etc...). [Why react.memo](https://dmitripavlutin.com/use-react-memo-wisely/)


bordalix:
@louisinger to fix the blinking image, we just needed to `mv unknown.png unknown.svg` (it was a svg file with the wrong extension).

But I see you did more than that here. Makes sense what you did.

PR #357, opened 2022-06-06
support ionio-artifact template type

louisinger:
This PR adds support for `ionio-artifact` template (argument in `marina.importTemplate`).

if the template argument in `importTemplates` is `ionio-artifact` marina will try to build a valid descritptor from the artifact functions. Using `toDescriptor` func from @ionio-lang/ionio.

@tiero please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "liquidjs-lib": "git+https://github.com/louisinger/liquidjs-lib.git#psbt-taproot-support"
   },
   "dependencies": {
+    "@ionio-lang/ionio": "^0.1.5",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "liquidjs-lib": "git+https://github.com/louisinger/liquidjs-lib.git#psbt-taproot-support"
   },
   "dependencies": {
+    "@ionio-lang/ionio": "^0.1.5",
```
tiero, line 26:
```suggestion
    "@ionio-lang/ionio": "^0.1.7",
```

PR #361, opened 2022-06-16
`getAccountInfo()` returns mainAccount data + add main account template

### LINKED ISSUE #355: `await marina.getAccountInfo()` returns undefine  for MainAccount
on MainAccount f I do `await marina.getAccountInfo()` will get `Uncaught Error: Account undefined not found`

_Originally posted by @tiero in https://github.com/vulpemventures/marina/issues/345#issuecomment-1145945204_

### LINKED ISSUE #367: Make sure descriptor returned in getAccountInfo matches the one expected by neutrino-elements

louisinger:
use the constant `elwpkh($mainAccount)` as template in getAccountInfo() result.

it closes #355 
it fixes #367 

@tiero please review


src/domain/account.ts
```diff
@@ -114,6 +114,7 @@ function createMainAccount(
       accountID: MainAccountID, // main account is unique
       masterXPub: data.masterXPub,
       isReady: true, // always true for main account
+      template: `elwpkh($${MainAccountID})`,
```
tiero, line 117: is the same uses as go-elements descriptor package?

tiero, line 117: mmm, we should return the actual descriptor no?

```suggestion
      template: `elwpkh(${data.masterXPub})`,
```

tiero, line 117: or then we need an additional field `descriptor` which uses the real thing we could import into Elements core for example?

```suggestion
      template: `elwpkh($${MainAccountID})`,
      descriptor: `elwpkh(${data.masterXPub})`
```

src/domain/account.ts
```diff
@@ -114,6 +114,7 @@ function createMainAccount(
       accountID: MainAccountID, // main account is unique
       masterXPub: data.masterXPub,
```
tiero, line 115: I think we shoudl give out xpub not zpubs

src/domain/account.ts
tiero, line 117:
```suggestion
      isReady: true, // always true for main account
      template: `elwpkh($${MainAccountID})`,
      descriptor: `elwpkh(${toXpub(data.masterXPub)})`,
```

tiero, line 117:
```suggestion
      accountID: MainAccountID, // main account is unique
      masterXPub: toXpub(data.masterXPub),
      isReady: true, // always true for main account
      template: `elwpkh($${MainAccountID})`,
```

tiero: is this the child xpub? ie. unhardened? @louisinger

louisinger:
UPDATE
* bumps to ldk 0.5.4 it closes #367 
* rename "covenantDescriptor" to "contractTemplate" according to new #366 `CustomScriptAccountData`

@tiero

PR #366, opened 2022-07-04
Update custom script identity

### LINKED ISSUE #365: Rethink usage of custom script account
A custom script account can be seen as a special account with a special derivation path, where every generated internal/external keypair is associated with a very specific contract of the "same family", ie. they all have the same template.

Given this vision, the first thing a user should do is to create a new account by importing a template.
He, then, could generate the address for a specific contract by specifying its constructor arguments.
The user should use `getNextAddress(contructorParams)` to substitute the special account's namespace with an external pubkey, while `getNextChangeAddress(constructoParams)` should be used to does the same but with an internal pubkey.
To make the user's life easier, he should be able to specify constructor arguments as a JSON object containing strings or numbers values. Marina should take care of translating these args into "raw" types.
For this reason, at the moment only Ionio artifacts should be supported as template type. Marina descriptors lack in the definition of constructor args types, forcing the user to provide "raw" types which is not ideal. They still could be used as user info since it's an easy way for him to read a contract, therefore they will be returned as account's address info.

The concept of `changeTemplate` should be dropped since it makes no sense for the account to support a totally different template for internal keypairs.

altafan:
Before this, the constructor params of a contract must be passed when importing the template into a custom script account. Now, they can be passed to within `getNext*Address` method.
Furthermore, with this the only accepted import template type becomes Ionio artifacts. Marina descriptors are returned just as address info to make it a bit easier for the user to read the contract.

NOTE: the `changeTemplate` field has been dropped since it became unused. 

Closes #365.

Please @tiero @louisinger @bordalix review this.


src/content/marina/marinaBroker.ts
```diff
@@ -410,32 +433,33 @@ export default class MarinaBroker extends Broker<keyof Marina> {
...
+          //   const nextChangeAddress =
+          //     (await watchIdentity.getNextChangeAddress()) as TaprootAddressInterface;
+          //   isSpendableByMarina ||= Object.values(
+          //     analyzeTapscriptTree(nextChangeAddress.TaprootAddressInterface)
+          //   ).some(autoSpendableLeaf);
+          // }
+
+          // await this.store.dispatchAsync(
+          //   setIsSpendableByMarina(this.selectedAccount, isSpendableByMarina)
+          // );
```
altafan, line 461: @louisinger help wanted on this:
I didn't get the meaning of `isSpenableByMarina`. If I'm not wrong it's used only for custom script accounts, but why? In theory, every custom account imported is indeed spendable by marina, isn't it?

I left the code commented since I'm not sure what to do with this logic, but it cannot stay here because it requires generating an address at import time (with no constructor args), which is not allowed anymore.

src/content/marina/marinaBroker.ts
```diff
@@ -216,17 +213,38 @@ export default class MarinaBroker extends Broker<keyof Marina> {
           const account = this.accountSelector(this.state);
           const net = selectNetwork(this.state);
           const watchOnlyIdentity = await account.getWatchIdentity(net);
-          const nextAddress = await watchOnlyIdentity.getNextAddress();
+          let nextAddress: AddressInterface;
+          if (account.type === AccountType.MainAccount) {
+            nextAddress = await watchOnlyIdentity.getNextAddress();
+          } else {
+            let constructorParams: Record<string, string | number> | undefined;
+            if (params) {
```
tiero, line 221:
```suggestion
            if (params && params.length === 1) {
```

src/content/marina/marinaBroker.ts
```diff
@@ -216,17 +213,38 @@ export default class MarinaBroker extends Broker<keyof Marina> {
...
+          }
           const net = selectNetwork(this.state);
           const xpub = await account.getWatchIdentity(net);
-          const nextChangeAddress = await xpub.getNextChangeAddress();
+          let nextChangeAddress: AddressInterface;
+          if (account.type === AccountType.MainAccount) {
+            nextChangeAddress = await xpub.getNextChangeAddress();
+          } else {
+            let constructorParams: Record<string, string | number> | undefined;
+            if (params) {
```
tiero, line 243:
```suggestion
            if (params && params.length > 0) {
```

src/content/marina/marinaBroker.ts
```diff
@@ -410,32 +433,33 @@ export default class MarinaBroker extends Broker<keyof Marina> {
               changeCovenant?.template
             )
           );
+          // TODO: clarifiy the usage of isSpendableByMarina and understand
+          // how to set that value in case it's needed.
```
tiero, line 437: It should me moved from template loading to address generation

src/domain/customscript-identity.ts
```diff
@@ -107,7 +111,7 @@ export class CustomScriptIdentityWatchOnly extends Identity implements IdentityI
   readonly masterBlindingKeyNode: Mnemonic['masterBlindingKeyNode'];
   readonly masterPubKeyNode: Mnemonic['masterPublicKeyNode'];
   readonly namespace: CustomScriptIdentityOpts['namespace'];
-  readonly covenant: CovenantDescriptors;
+  readonly covenant: ContractTemplate;
```
tiero, line 114: It's fun, cause we can have a Script that has NO covenant insider (ie. no introspection or timelocks in the `require` field)

src/domain/customscript-identity.ts
```diff
@@ -171,35 +162,76 @@ export class CustomScriptIdentityWatchOnly extends Identity implements IdentityI
...
+        case PrimitiveType.DataSignature:
+        case PrimitiveType.PublicKey:
+        case PrimitiveType.XOnlyPublicKey: {
+          if (typeof param === 'string') {
+            return Buffer.from(param, 'hex');
+          }
+          const buf = Buffer.alloc(8);
+          buf.writeBigUInt64LE(BigInt(param));
+          return buf;
+        }
```
tiero, line 198: Why you doing this? you are encoding a number when XOnlyPublicKey? 🤔

tiero: All types are handled by Ionio already, in the case of 64bit numbers, at the moment the idea is to give them passed as `bytes`, so the user (outside marina) should care about doing the right encoding and simply passing to us as 0x prefixed hex string. Ionio will know hot to handle from there

test/customscript.spec.ts
```diff
@@ -85,110 +67,97 @@ describe('CustomScriptIdentity', () => {
...
+    await faucet(addr.confidentialAddress, 0.01);
+
+    const [utxo] = await getUnspents(addr);
+
+    const instance = addr.contract.from(utxo.txid, utxo.vout, utxo.prevout);
+    const signer: Signer = {
+      signTransaction: async (psetBase64: string): Promise<string> => {
+        return random.signPset(psetBase64);
+      },
+    };
```
altafan, line 133: @tiero something's wrong with this Signer interface: the error retuned by broadcastTx suggests that the signature in the input witness is not valid. Any idea?

tiero: need to be investigate how signPset is producing the signature, has been updated to support taproot signing (hasForWitnessv1 etc..)?

tiero: Seems there is some taproot logic there, so it should work. 

Can you inspect how the witness is for that covenant input you trying to unlock?

ideally, the end result would be to update the `sig` made with `ecc.signSchnorr` inside the input and return that psbt

```ts
ptx.updateInput(index, {
          tapScriptSig: [
            {
              leafHash,
              pubkey: keyPair.publicKey.slice(1),
              signature: sig,
            },
          ],
        });

```

tiero: maybe @louisinger can help, that logic in signPset is huge and poorly documented as of now (plus does a lot of magic tricks, so would be good occasion to simplify and strip it down)

test/customscript.spec.ts
```diff
@@ -85,110 +67,88 @@ describe('CustomScriptIdentity', () => {
...
-        },
-      ],
-      utxos,
-      networks.regtest.assetHash,
-      () => addr.confidentialAddress,
-      'regtest'
-    );
+  test('should be able to instatiate a contract identity, fund the contract and spend those funds', async () => {
+    const template = JSON.stringify(transferWithCaptchaArtifact);
+    const random = makeRandomCustomScriptIdentity(template);
```
tiero, line 108: where you passing the constructor arguments? ie. the sum  (3+4=7?) and the pubkey?

test/customscript.spec.ts
louisinger: related to https://github.com/altafan/marina/pull/1

src/content/marina/marinaBroker.ts
```diff
@@ -394,48 +420,31 @@ export default class MarinaBroker extends Broker<keyof Marina> {
...
-            isSpendableByMarina ||= Object.values(
-              analyzeTapscriptTree(nextChangeAddress.TaprootAddressInterface)
-            ).some(autoSpendableLeaf);
-          }
+          const artifact: Artifact = JSON.parse(contract!.template);
+          const isSpendableByMarina = artifact.functions.every(({ functionInputs, require }) => {
+            return (
+              functionInputs.length === 1 &&
+              functionInputs[0].type === PrimitiveType.Signature &&
+              (!require || require.length === 0)
```
louisinger, line 440: nit: require could lead to conflict with the JS require :thinking:

test/customscript.spec.ts
```diff
@@ -85,110 +67,88 @@ describe('CustomScriptIdentity', () => {
...
+    expect(random.contract.namespace).toBe(TEST_NAMESPACE);
+    expect(random.contract.template).toBeDefined();
+
+    const addr = await random.getNextAddress({ sum: 7 });
```
altafan, line 111: > where you passing the constructor arguments? ie. the sum (3+4=7?) and the pubkey?

@tiero here in `getNextAddress`

package.json
```diff
@@ -36,7 +36,7 @@
     "decimal.js": "^10.2.1",
     "formik": "^2.2.9",
     "google-protobuf": "^3.15.8",
-    "ldk": "^0.5.2",
+    "ldk": "^0.5.3",
     "lodash.debounce": "^4.0.8",
     "lottie-web": "^5.7.8",
     "marina-provider": "^1.7.0",
```
bordalix, line 42: Should it be 1.7.1?

tiero: Yep! It must be to have new interface

tiero: https://github.com/vulpemventures/marina-provider/commit/48c705d3e81eec01551d487d4461783ccec78aa0

bordalix:
Changes are needed on src/inject/marina/provider.ts

```
--- a/src/inject/marina/provider.ts
+++ b/src/inject/marina/provider.ts
@@ -66,12 +66,12 @@ export default class Marina extends WindowProxy<keyof MarinaProvider> implements
     return this.proxy('getAddresses', [ids]);
   }
 
-  getNextAddress(): Promise<AddressInterface> {
-    return this.proxy('getNextAddress', []);
+  getNextAddress(constructorParams?: any): Promise<AddressInterface> {
+    return this.proxy('getNextAddress', [constructorParams]);
   }
 
-  getNextChangeAddress(): Promise<AddressInterface> {
-    return this.proxy('getNextChangeAddress', []);
+  getNextChangeAddress(constructorParams?: any): Promise<AddressInterface> {
+    return this.proxy('getNextChangeAddress', [constructorParams]);
   }
```

src/content/marina/marinaBroker.ts
tiero: @altafan

bordalix:
Here are my notes from our call:

Whenever we use `account.getWatchIdentity(network)`, the account is restored.

In this process of restoring, all previous addresses are derived.

Since we don’t know which constructorParams were used in previous addresses, we can’t derive the previous addresses of a custom script account.

LDK code for restoring from state

```
export function restorerFromState<R extends IdentityInterface>(
  identity: R
): Restorer<StateRestorerOpts, R> {
  return async ({ lastUsedExternalIndex, lastUsedInternalIndex }) => {
    const promises = [];

    if (lastUsedExternalIndex !== undefined) {
      for (let i = 0; i <= lastUsedExternalIndex; i++) {
        const promise = identity.getNextAddress();
        promises.push(promise);
      }
    }

    if (lastUsedInternalIndex !== undefined) {
      for (let i = 0; i <= lastUsedInternalIndex; i++) {
        const promise = identity.getNextChangeAddress();
        promises.push(promise);
      }
    }

    await Promise.all(promises);

    return identity;
  };
}
```

One solution would be to store all constructorParams used previously, and pass them to the restorer, which implies a **change on LDK** and **storing blobs of information**. This seems impractical and cumbersome.

Also, I don’t see how we could find the public keys used on previous contracts, the keys that own the USDT UTXO. We would need to store them also on the blob of information.

Another possibility discussed was to not increase the lastUsedInternalIndex for customScriptAccount. This would make customScriptAccounts having just one contract address, the last contract used. But again, we would lose the pubkeys used previously (that own the USDT UTXO). Not sure if this could work either.

tiero:
What we discussed back in the days, was having an API such as `importContract`(s) where the user must load constructor params and tells marina either a) which `address`/`scriptPubKey` they belongs to b) the exact index/derivation path of the pubkey present in the script  

Of course the former is best, latter is basically a leaky abstraction.

tiero:
> and storing blobs of information

yes, I think topics like "contract metadata stores" offered by explorer would be key to this, for now, we should assume user/web apps need to take care of backup contract address and their constructor parameters.

tiero:
> The "custom" restore doesn't need to call the identity's getNext*Address() but instead it has to manually derive the pubkeys and "translate" them into p2wpkh addresses. For now, it's fine to assume that the addresses generated by keys used as contracts' constructor params are always of type p2wpkh (segwit v0).

Not clear to me this comment: if we are talking about custom contracts, most likely they have more constructor parameters than just the public key.

If user wants to load into a custom account which unlock funds with only marina pubkey in a single script path spend  ie. `eltr(H_POINT, { elwpkh($marinaPubKey) }`, I think we should treat in the same way of other custom script account, having user/webapp to import the contract template, parameters etc...

louisinger:
> > and storing blobs of information
> 
> yes, I think topics like "contract metadata stores" offered by explorer would be key to this, for now, we should assume user/web apps need to take care of backup contract address and their realtive constructor parameters.

Instead of storing the whole constructorParams, Marina could only save the script tree merkle root. For "Watch" processes it lets to re-create the addresses (using internal pub key + the root) and also let us to check "validity" of constructorParams if later an app needs to send them to Marina (I'm thinking about signing for instance).

bordalix:
> > > and storing blobs of information
> > 
> > 
> > yes, I think topics like "contract metadata stores" offered by explorer would be key to this, for now, we should assume user/web apps need to take care of backup contract address and their realtive constructor parameters.
> 
> Instead of storing the constructorParams object, Marina could only save the script tree merkle root. For "Watch" processes it lets to re-create the addresses (using internal pub key + the root) and also let us to check "validity" of constructorParams if later an app needs to send them to Marina because we can check if computed root equals saved root (I'm thinking about signing for instance).

Will this implies a change on LDK?

tiero:
btw I thought we were talking about restoring a mnemonic, actually I understood from @altafan that it'a about the Custom Script Identity and his Restorer implemented in Marina, which does not have yet support for storing constructorInputs, which are necessary to derive correct addresses when destroying/recreating the LDK instance (common in frontends, where you instantiate on every page load)

I guess we should not really change restorer & updater, just implement the right "RestorerFromRedux" in marina to source right addresses to the updater

bordalix:
Additional note:

Code as is, makes change address a contract address.
Do we really want it to be? I don't think so.

tiero:
> Code as is, makes change address a contract address. Do we really want it to be? I don't think so.

We are not going to use the `getNextChangeAddress` in our apps, that being said there is a way to source different templates for the internal chain (ie. change)

The best way would be to pass a basic `eltr($namespace, {})` or  `eltr(H_POINT, {elpwpkh($namespace)})` as change template no? so it's auto-spendable by marina UI cc/ @louisinger

bordalix:
> > Code as is, makes change address a contract address. Do we really want it to be? I don't think so.
> 
> We are not going to use the `getNextChangeAddress` in our apps, that being said there is a way to source different templates for the internal chain (ie. change)
> 
> The best way would be to pass a basic `eltr($namespace, {})` or `eltr(H_POINT, {elpwpkh($namespace)})` as change template no? so it's auto-spendable by marina UI cc/ @louisinger

I was referring to the change address for the funding tx, the get the collateral change.
I guess I will use a change address from the main account, right?

src/domain/customscript-identity.ts
```diff
@@ -546,3 +546,43 @@ function validateTemplate(template: string): boolean {
...
+    lastUsedExternalIndex,
+    lastUsedInternalIndex,
+    customParamsByIndex,
+    customParamsByChangeIndex,
+  }) => {
+    const promises = [];
+
+    if (lastUsedExternalIndex !== undefined) {
+      for (let i = 0; i <= lastUsedExternalIndex; i++) {
+        const params = customParamsByIndex[lastUsedExternalIndex];
```
bordalix, line 570: `const params = customParamsByIndex[i];` instead, right? cc @altafan

src/domain/customscript-identity.ts
```diff
@@ -546,3 +546,43 @@ function validateTemplate(template: string): boolean {
...
+        const promise = identity.getNextAddress(params);
+        promises.push(promise);
+      }
+    }
+
+    if (lastUsedInternalIndex !== undefined) {
+      for (let i = 0; i <= lastUsedInternalIndex; i++) {
+        const params = customParamsByChangeIndex[lastUsedExternalIndex];
```
bordalix, line 578: `const params = customParamsByChangeIndex[i];` instead, right? cc @altafan

bordalix (changes requested):

bordalix:
In the decoded transaction before broadcasting, the decoded script for the covenant (vout[0]) seems weird:

```
{
  "txid": "3696e8265674ceb8aed12cf38128ef18262825d82e9854b039cb828f00041112",
  "hash": "ab8c265167f1e0c3de4d0da7a34bc06b452c05a1575a46247bd5e184e5a77808",
  "wtxid": "ab8c265167f1e0c3de4d0da7a34bc06b452c05a1575a46247bd5e184e5a77808",
  "withash": "6e055710b822e0ffc0b7214daa8a7b7a41128e484818cc4bb1994715e4cdf83a",
  "version": 2,
  "size": 7061,
  "vsize": 2349,
  "weight": 9395,
  "locktime": 0,
  "vin": [
    {
      "txid": "3fba68b301621361274daa833e786d8e6a88b20bb968e35e2f3e94b74ac1ee17",
      "vout": 0,
      "scriptSig": {
        "asm": "",
        "hex": ""
      },
      "is_pegin": false,
      "sequence": 4294967295,
      "txinwitness": [
        "3045022100e1a7bfb8aea26ddbc8b4cee24d05fdc8ad178010e74be83b746a65ba381a063e02201154b10aed0a6520a558461a2d207e941ff2ade0a18f49229362eeaff85dd72101",
        "03d49a04085d5091bace5bfc508b105bed1d4d34841e44955b8436b5ab9c171870"
      ]
    },
    {
      "txid": "b432cda9cb7e23fd2453ac4879d1e6a48dead02032e5440a9021fd10c26a4228",
      "vout": 3,
      "scriptSig": {
        "asm": "",
        "hex": ""
      },
      "is_pegin": false,
      "sequence": 4294967295,
      "txinwitness": [
        "304402206ce9a9b4d1ed2663857155092efb8601c27f2bf37ad85faf406bda25213f77b202201ada9487ee0513637687c672f4784b1c4bf7825e768fec54d8d9912e0463a22601",
        "02a3d89442c53fa319737bce93e0408d00f817f992a276fea1a2aa0ffbdc4a8a76"
      ],
      "issuance": {
        "assetBlindingNonce": "d5ce48ea9a18818d7731f2e925033a47732a8e4c745b45dc65369cd6aa130a7a",
        "assetEntropy": "367bd63bca062ce1ee00ac8a7308064fa5cf9c63f324f28e4c69b325809491b0",
        "isreissuance": true,
        "asset": "04e28b858766654440399712cfcd49bcfa512971b7e79cd4029dbb23d18cd568",
        "assetamount": 100.00000000
      }
    }
  ],
  "vout": [
    {
      "value": 0.01034425,
      "asset": "144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49",
      "commitmentnonce": "",
      "commitmentnonce_fully_valid": false,
      "n": 0,
      "scriptPubKey": {
        "asm": "OP_DEPTH OP_FROMALTSTACK OP_2ROT 70713033366463636e763730616e6e6c6a6b776a6b63326876676c636c7273706876776e6b6877666a7a7932306b6d7763 6d75766e64307377716d73753666786b386b777979336574663366666d366777656e786867636d3273323764687a3237766c OP_ROLL OP_PICK [error]",
        "hex": "746c713170713033366463636e763730616e6e6c6a6b776a6b63326876676c636c7273706876776e6b6877666a7a7932306b6d7763326d75766e64307377716d73753666786b386b777979336574663366666d366777656e786867636d3273323764687a3237766c7a793875747967337a36733061776a6c35",
        "type": "nonstandard"
      }
    },
    {
      "value-minimum": 0.00000001,
      "value-maximum": 687.19476736,
      "ct-exponent": 0,
      "ct-bits": 36,
      "surjectionproof": "0300072542e3a4cd7efebaa33ce37e249bad5608f1ecdd4135184856b412e40614591ddb9d7327e686101f1cf243e479f8d6e5f61c526e0bc169a5976ae7b3339ad365ddcee5425e849a5637d36e8eebf22dc1f7b16972e928470aa7b90156b53838f6aed0d88103997c5d9c9e11a4cf7f27017f1f8d23584644d700d5f1964f87b65c",
      "valuecommitment": "08bb43ce31ae291c845680b46d17cadb0570bc2570f5282f8b419853f3b0eac97a",
      "assetcommitment": "0a9bed717561ef16ed8e6f78fe4f2690af0d564bd2c9793cafc36abe6e20605a06",
      "commitmentnonce": "02958348c3b043169fb250e4539c69d9a3787d62c7d193b24b981ab8810295b8e0",
      "commitmentnonce_fully_valid": true,
      "n": 1,
      "scriptPubKey": {
        "asm": "OP_DEPTH OP_FROMALTSTACK OP_2ROT 7171643976736a726c67377063726d333368336d727179343461706173386663676c6668617a34367368757933707a6c74 [error]",
        "hex": "746c71317171643976736a726c67377063726d333368336d727179343461706173386663676c6668617a34367368757933707a6c74346a65653064796a7766703968786e326b7672307674616d383037677a38776e71636d38373761687275776571717a746c",
        "type": "nonstandard"
      }
    },
    {
      "value": 100.00000000,
      "asset": "04e28b858766654440399712cfcd49bcfa512971b7e79cd4029dbb23d18cd568",
      "commitmentnonce": "",
      "commitmentnonce_fully_valid": false,
      "n": 2,
      "scriptPubKey": {
        "asm": "1 b5f070370e6926b1ece212395a6294ef4876666ba31b5415e6dc4af33e221f8b",
        "hex": "5120b5f070370e6926b1ece212395a6294ef4876666ba31b5415e6dc4af33e221f8b",
        "reqSigs": 1,
        "type": "witness_v1_taproot",
        "addresses": [
          "tex1pkhc8qdcwdyntrm8zzgu45c55aay8vent5vd4g90xm390x03zr79s4lkkgj"
        ]
      }
    },
    {
      "value-minimum": 0.00000001,
      "value-maximum": 687.19476736,
      "ct-exponent": 0,
      "ct-bits": 36,
      "surjectionproof": "03000734a7249ab384e65b6ff4ad10cb8f12482cd7eec1613db76ebbb73d57acf0a2f045cbe881e5020f8ca50bb00f8e4dc8107e28ef2d40f64df81e68f27e0638f7330932e559f8e07f81dded802ea9a8ccfae23c72d6ea00eb0243fedcd7f11df561881c9bfe92c19cefaf5d7cf9923d33c26f27b62cada8c2b676292e9596e5503a",
      "valuecommitment": "08faf7e623d5a0dfe40140767f4a5afd4f30f5ae53b85ae8485c2db4c535a01772",
      "assetcommitment": "0bdf6e4f614893da4df7d57495042a60d553e83d63cc3408e78945222c8855e552",
      "commitmentnonce": "026a15ec7d087f49e76e2fd018bdbb57167eef6440a79ce9c2128ff6c1ab04eb38",
      "commitmentnonce_fully_valid": true,
      "n": 3,
      "scriptPubKey": {
        "asm": "0 5af91eef0239003c4c1cc4efaf4270327595e76b",
        "hex": "00145af91eef0239003c4c1cc4efaf4270327595e76b",
        "reqSigs": 1,
        "type": "witness_v0_keyhash",
        "addresses": [
          "tex1qttu3amcz8yqrcnqucnh67snsxf6etemthl4l0k"
        ]
      }
    },
    {
      "value": 0.00000500,
      "asset": "144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49",
      "commitmentnonce": "",
      "commitmentnonce_fully_valid": false,
      "n": 4,
      "scriptPubKey": {
        "asm": "",
        "hex": "",
        "type": "fee"
      }
    }
  ]
}
```
cc @louisinger @altafan

bordalix:
The marina.getNextAddress() object for a custom script account as a lot of information.

Do we really need all this?

https://gist.github.com/bordalix/1632b849280c3ba29d71147a5a434a3d

cc @altafan @louisinger

tiero:
One thing suspicious I see is that both vout0 and vout1 scripts are decoded as `nonstandard` script types and have some error on the `asm` field, so they seem corrupt.

The vout2 instead it should theoretically be a segwit v0 id fetched from `mainAccount`

tiero:
it seems vou1 and vout2 are fixed (ie. using mainAccount)

Remains problem of the added covenant script as vout0 which is decoded badly

```
{
  "txid": "f49a8923458edf8d3803c55d16b02172b7e2c6cd95c3c6aadac7120d0bc06c72",
  "hash": "9ce47467220a7aba9eac2750ec4c73353c5e29a6dfb85114ed75a5eab86dd743",
  "wtxid": "9ce47467220a7aba9eac2750ec4c73353c5e29a6dfb85114ed75a5eab86dd743",
  "withash": "2ceff7b8a27fc9f47152575a345bcb4b26f82da0f9eb5af0d6406b36316bc2f2",
  "version": 2,
  "size": 6968,
  "vsize": 2257,
  "weight": 9026,
  "locktime": 0,
  "vin": [
    {
      "txid": "3fba68b301621361274daa833e786d8e6a88b20bb968e35e2f3e94b74ac1ee17",
      "vout": 0,
      "scriptSig": {
        "asm": "",
        "hex": ""
      },
      "is_pegin": false,
      "sequence": 4294967295,
      "txinwitness": [
        "3044022023037671cb67905f83a9ad4d84cf01d57125113ebd71ac2d5c8ee65add0107d3022043f5c261ad74d2b5ff613666ef79ca7ac80ffeedd0afa09044e0a5d5c42d254101",
        "03d49a04085d5091bace5bfc508b105bed1d4d34841e44955b8436b5ab9c171870"
      ]
    },
    {
      "txid": "b432cda9cb7e23fd2453ac4879d1e6a48dead02032e5440a9021fd10c26a4228",
      "vout": 3,
      "scriptSig": {
        "asm": "",
        "hex": ""
      },
      "is_pegin": false,
      "sequence": 4294967295,
      "txinwitness": [
        "30440220163bcfbd042fa7ec5b6076cfd734e05321f59239106da68035caa0dd45d22092022001503458ac5986fdda3b504085bf9126bfd2d5fc00b9e2c52197488405c54cb601",
        "02a3d89442c53fa319737bce93e0408d00f817f992a276fea1a2aa0ffbdc4a8a76"
      ],
      "issuance": {
        "assetBlindingNonce": "d5ce48ea9a18818d7731f2e925033a47732a8e4c745b45dc65369cd6aa130a7a",
        "assetEntropy": "367bd63bca062ce1ee00ac8a7308064fa5cf9c63f324f28e4c69b325809491b0",
        "isreissuance": true,
        "asset": "04e28b858766654440399712cfcd49bcfa512971b7e79cd4029dbb23d18cd568",
        "assetamount": 100.00000000
      }
    }
  ],
  "vout": [
    {
      "value": 0.01044015,
      "asset": "144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49",
      "commitmentnonce": "",
      "commitmentnonce_fully_valid": false,
      "n": 0,
      "scriptPubKey": {
        "asm": "OP_DEPTH OP_FROMALTSTACK OP_2ROT 7071306b6e6379796e64363667743865647463797a66657638666336656163656e3761643068333966616a646377726a76 OP_VERNOTIF OP_VERNOTIF OP_IFDUP 39646a376e6b7830617032633067797763666a657168773739777678716761376664396a3674717a737a716674613965617134647a30 OP_2ROT OP_IFDUP [error]",
        "hex": "746c71317071306b6e6379796e64363667743865647463797a66657638666336656163656e3761643068333966616a646377726a766666733639646a376e6b7830617032633067797763666a657168773739777678716761376664396a3674717a737a716674613965617134647a3071733963656861333376",
        "type": "nonstandard"
      }
    },
    {
      "value-minimum": 0.00000001,
      "value-maximum": 687.19476736,
      "ct-exponent": 0,
      "ct-bits": 36,
      "surjectionproof": "03000747a14e6d3ddcb5bf3c629bbde066c69ac4050aa9c61467a5c419a621a2faa59885d370ef94fe96a05006188ce5b0d7ae2c4346a61d53e0e14942fd31d029e12199a19a1040ec32d4980283276dc47fedf9c775d85784dbea8382dfd881a8fef2ad126ddd85b60ef1ccd33bc5924f41b9b0f3f18598cef3982e113b60bf0a6679",
      "valuecommitment": "098d78e5f607ebbe1afeea0f69ebb030a43567c5cf12370148b014ab6d31401f88",
      "assetcommitment": "0bd87aec440a09efa93c2a3749dd903789be7f68364cb130fc25c272eb7f556372",
      "commitmentnonce": "03d5effe22b2c5f3aa20244c4e279e7d26cb5c60b33e8259d6a343c1d0b0c3f0c9",
      "commitmentnonce_fully_valid": true,
      "n": 1,
      "scriptPubKey": {
        "asm": "0 05a8787467401dc09e3032baecde4b8ac9e6c0bb",
        "hex": "001405a8787467401dc09e3032baecde4b8ac9e6c0bb",
        "reqSigs": 1,
        "type": "witness_v0_keyhash",
        "addresses": [
          "tex1qqk58sar8gqwup83sx2awehjt3ty7ds9m3v2jwn"
        ]
      }
    },
    {
      "value": 100.00000000,
      "asset": "04e28b858766654440399712cfcd49bcfa512971b7e79cd4029dbb23d18cd568",
      "commitmentnonce": "",
      "commitmentnonce_fully_valid": false,
      "n": 2,
      "scriptPubKey": {
        "asm": "0 f39852f219f9b932f9c094b447a4d7b1f114172a",
        "hex": "0014f39852f219f9b932f9c094b447a4d7b1f114172a",
        "reqSigs": 1,
        "type": "witness_v0_keyhash",
        "addresses": [
          "tex1q7wv99uselxun97wqjj6y0fxhk8c3g9e2e8jlcp"
        ]
      }
    },
    {
      "value-minimum": 0.00000001,
      "value-maximum": 687.19476736,
      "ct-exponent": 0,
      "ct-bits": 36,
      "surjectionproof": "030007d82bc29a11850d81d50efdfb5232642b7a08d8be31bd75ec2263f3b7b2344e5413fe127896813e040dae5d80928a3c4ff0191538650b329bdb317aee125853d2bee130187a3cddee66d9352b03782071008cb0e82e30af5805064f11ed934e99b4c8ab15e4ed1f1995444d0f5c3fd20a77758e4bed13ba1cf2fb6a782cae8b50",
      "valuecommitment": "092d33ef138272a97632ebd7847669bc8b8016c88d05923a807566105858be4c75",
      "assetcommitment": "0ba139bcf20bf85be26f7d7255ade7314f6e3083f733f6047939f9b877860696ff",
      "commitmentnonce": "03e84a0c2162d1cdb042ff9e899d8862d0b98cc2e47bada68d38106d03d3c356e2",
      "commitmentnonce_fully_valid": true,
      "n": 3,
      "scriptPubKey": {
        "asm": "0 5af91eef0239003c4c1cc4efaf4270327595e76b",
        "hex": "00145af91eef0239003c4c1cc4efaf4270327595e76b",
        "reqSigs": 1,
        "type": "witness_v0_keyhash",
        "addresses": [
          "tex1qttu3amcz8yqrcnqucnh67snsxf6etemthl4l0k"
        ]
      }
    },
    {
      "value": 0.00000500,
      "asset": "144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49",
      "commitmentnonce": "",
      "commitmentnonce_fully_valid": false,
      "n": 4,
      "scriptPubKey": {
        "asm": "",
        "hex": "",
        "type": "fee"
      }
    }
  ]
}
```

bordalix:
All problems were on my side.
It's fixed, and we have the first borrow contract on testnet:
https://blockstream.info/liquidtestnet/tx/906b58fe131c19f2eef2222cab503508fb0969a88a2bcad9974eb74fdd3a59b5

tiero:
> custom script account as a lot of information.

I think the `descriptor`, `contract` and constructor params are more than enough. Not sure how useful is the `taprootTree`

re: `contract` I noticed that inside the `asm` field the $template strings are replaced, but not in the `require` field. Is that wanted?

src/domain/customscript-identity.ts
```diff
@@ -79,11 +73,10 @@ function asTaprootAddressInterface(
     confidentialAddress: extended.confidentialAddress,
     blindingPrivateKey: extended.blindingPrivateKey,
     derivationPath: extended.derivationPath,
-    taprootHashTree: extended.result.taprootHashTree,
-    taprootInternalKey: extended.result.taprootInternalKey,
```
tiero, line 78: maybe this should be kept?

tiero: or can be extracted from the descriptor?

louisinger: I would drop taproot members, the field `contract` (Ionio contract) should be enough to compute "taproot witness stack" for signing step

tiero: right, but Ionio Artifacts currently does not expose the internal PublicKey used in the compilation of the script.

But point taken, will expose a new field in the artifact then to retrieve it from there https://github.com/ionio-lang/ionio/issues/13

tiero:
If you create an account for a network, import a template with constructor params ie. Liquid and then switch to Testnet from settings you get an error.

image

We should handle this, maybe the accounts should be segregated for network? @altafan @louisinger

src/domain/customscript-identity.ts
altafan: These fields have been removed just from the type returned to the user, they should be cached by the identity still.

louisinger:
> We should handle this, maybe the accounts should be segregated for network? @altafan @louisinger

The error is happening everytime Marina is using the deepRestorer on custom accounts _with constructor params_(including network switch). On that "custom w/ param accounts", we can't apply the bip32-like restoration under esplora because each address needs a set of parameters. 

My idea was to add them to the `restorerFromEsplora` to recompute addresses but the goal of that restorer is to check the next addresses on chain, and we fail to generate them cause we don't have any parameter for the next addresses.

--
maybe "deepRestorer" does not make sense on that type of account. Either we have our custom params and thus we know the last index or we don't have params so even with the index we can't regenerate the addresses.
@tiero @altafan

tiero:
> we don't have params so even with the index we can't regenerate the addresses

That's fine as condition (ie. restore mainAccount always), but we should try restoration of custom accounts where we DO have cached constructorParams eventually, but without discovery, just derive the ones we have cached

tiero:
Is the problem you reported fixed with the latest version? @bordalix

bordalix:
Yes.

@tiero

tiero:
any idea how to make pipeline green @altafan @louisinger

src/domain/customscript-identity.ts
```diff
@@ -374,53 +373,59 @@ export class CustomScriptIdentity
...
+                    bip341.findScriptPath(tree, leafHash)
+                  );
+
+                pset.data.inputs[index].tapLeafScript = pset.data.inputs[index].tapLeafScript?.slice(1); // clear tapLeafScript first (we'll overwrite it)
+                pset.updateInput(index, {
+                  tapLeafScript: [
+                    {
+                      leafVersion: 0xc4, // elements tapscript version
+                      script: Buffer.from(leafScript, 'hex'),
+                      controlBlock: taprootSignScriptStack[1],
```
tiero, line 410: not really understand this code: ideally user already updated input before signing?

louisinger: This is in case there is no leaf on the input (and it's not key path). In that case marina will try to add the "auto-spendable" leaf if any exist. we need it for the finalize step.

PR #378, opened 2022-08-18
Bug fix and improvement on marina events

### LINKED ISSUE #377: NEW_UTXO and SPENT_UTXO  events should be fired for all accounts and return the accountID
```ts
marina.on('NEW_UTXO', (utxo, accountId) => {})
```

bordalix:
Closes #377

**Bug fixes:**

- fix bug where only the main account was emitting events on `NEW_TX`, `NEW_UTXO` and `SPENT_UTXO`

**Improvements:**

- marina events (see list above) now returns the accountID in the payload

**Breaking changes (marina events):**

- `marina.on('NEW_TX', (tx) => {})` is now `marina.on('NEW_TX', ({ tx, accountID }) => {})`
- `marina.on('NEW_UTXO', (utxo) => {})` is now `marina.on('NEW_UTXO', ({ utxo, accountID }) => {})`
- `marina.on('SPENT_UTXO', (utxo) => {})` is now `marina.on('SPENT_UTXO', ({ utxo, accountID }) => {})`

**Worth mentioning:**

- adds accountID to console.debug when reporting utxos fetched from explorer

@tiero @louisinger please review.

### DIFF
--- a/src/application/utils/marina-event.ts
+++ b/src/application/utils/marina-event.ts
@@ -7,9 +7,9 @@ export interface MarinaEvent<P extends any> {
   payload: P;
 }
 
-export type NewUtxoMarinaEvent = MarinaEvent<UnblindedOutput>;
-export type SpentUtxoMarinaEvent = MarinaEvent<Outpoint>;
-export type NewTxMarinaEvent = MarinaEvent<TxDisplayInterface>;
+export type NewUtxoMarinaEvent = MarinaEvent<{ utxo: UnblindedOutput; accountID: string }>;
+export type SpentUtxoMarinaEvent = MarinaEvent<{ utxo: Outpoint; accountID: string }>;
+export type NewTxMarinaEvent = MarinaEvent<{ tx: TxDisplayInterface; accountID: string }>;
 export type EnabledMarinaEvent = MarinaEvent<{ network: NetworkString; hostname: string }>;
 export type DisabledMarinaEvent = MarinaEvent<{ network: NetworkString; hostname: string }>;
 export type NetworkMarinaEvent = MarinaEvent<NetworkString>;
@@ -17,7 +17,8 @@ export type NetworkMarinaEvent = MarinaEvent<NetworkString>;
 // compare tx history states and return marina events
 export function compareTxsHistoryState(
   oldState: TxsHistory,
-  newState: TxsHistory
+  newState: TxsHistory,
+  accountID: string
 ): NewTxMarinaEvent[] {
   const events: NewTxMarinaEvent[] = [];
   const newEntries = Object.entries(newState);
@@ -25,7 +26,7 @@ export function compareTxsHistoryState(
 
   for (const [txID, tx] of newEntries) {
     if (oldTxIDs.includes(txID)) continue;
-    events.push({ type: 'NEW_TX', payload: tx });
+    events.push({ type: 'NEW_TX', payload: { tx, accountID } });
   }
 
   return events;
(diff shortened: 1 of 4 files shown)


tiero:
If we pass an additional parameter, instead of changing it to be an object, we should be able to make it not breaking change?

`marina.on('NEW_TX', (tx, accountID?) => {})`

bordalix:
>  If we pass an additional parameter, instead of changing it to be an object, we should be able to make it not breaking change?
> 
> `marina.on('NEW_TX', (tx, accountID?) => {})`

@louisinger help me out here please, but I think this will implicate a profound change on Marina.
Today, the payload for the events is passed between components as a single object (which gets stringified).
@tiero what you ask for would imply to have multiple objects being passed through.

bordalix:
@tiero to avoid more breaking changes, all payloads should be on the form `{ network }` instead of `network`.
This way we can add more attributes without breaking previous code.

tiero:
Agreed. I don't see any other way.

I woudl eventually make a typed interface: 

```ts
interface MarinaEvent {
  accountID: string,
  payload: any
}
```

louisinger:
> If we pass an additional parameter, instead of changing it to be an object, we should be able to make it not breaking change?
> 
> `marina.on('NEW_TX', (tx, accountID?) => {})`

it means update the [MarinaProvider interface](https://github.com/vulpemventures/marina-provider/blob/master/src/provider.ts#L36). 

But anyway not sure how marina background broker can handle this. if user has passed a 1-param func previously, `func(a, undefined)` statement will throw an error anyway. I personally prefer passing an object like joao proposes...

PS: the only way to avoid breaking changes is to return the accountID **inside the current payload**, so in NEW_UTXO we return `{ ...utxo, accountID }` instead of `utxo`. The drawback is `utxo` type don't have the accountID member so it's a kind of "hidden" member from a TS point of view.

PPS: note that due to `payload` typing which depends on the event type, the "breaking change" is a silent one -> it means that linter and TS checks won't see the error (even if a good code should check the type returned by marina)

src/application/utils/marina-event.ts
```diff
@@ -7,25 +7,26 @@ export interface MarinaEvent<P extends any> {
   payload: P;
 }
 
-export type NewUtxoMarinaEvent = MarinaEvent<UnblindedOutput>;
-export type SpentUtxoMarinaEvent = MarinaEvent<Outpoint>;
-export type NewTxMarinaEvent = MarinaEvent<TxDisplayInterface>;
+export type NewUtxoMarinaEvent = MarinaEvent<{ utxo: UnblindedOutput; accountID: string }>;
+export type SpentUtxoMarinaEvent = MarinaEvent<{ utxo: Outpoint; accountID: string }>;
```
louisinger, line 11: nit: I would replace `Outpoint` by `UnblindedOutput` (we pass the whole utxo so user will be able to access all members)

louisinger (approved):
utACK

bordalix:
Final merge was:

```typescript
interface MarinaEvent {
    accountID?: string
    data: any
}
```

With **data** type depending on event type:

```typescript
marina.on('NEW_TX', ({ accountID, data }) => {}) // data: Transaction
marina.on('NEW_UTXO', ({ accountID, data }) => {}) // data: UnblindedOutput
marina.on('SPENT_UTXO', ({ accountID, data }) => {}) // data: UnblindedOutput
marina.on('ENABLED', ({ data }) => {}) // data: { hostname: string; network: NetworkString }
marina.on('DISABLED', ({ data }) => {}) // data: { hostname: string; network: NetworkString }
marina.on('NETWORK', ({ data }) => {}) // data: NetworkString
```

Updated on https://docs.vulpem.com/marina/api/#marina-events

PR #392, opened 2022-08-24
Use `ElectrsBatchServer` from ldk 0.5.6

### LINKED ISSUE #381: Make Mempool default explorer

### LINKED ISSUE #382: Use new electrs batch server to load utxos & txs
https://electrs-batch.vulpem.com/status
https://electrs-batch-testnet.vulpem.com/status

louisinger:
**This PR uses the new `ChainAPI` abstraction to update the background updaters fetching and unblinding unspents and transactions.**

- bump ldk to 0.5.6
- add `batchServerURL` to `ExplorerURLs` data (already stored in reducer).
- `batchServerURL` is optional, not set on regtest.
- replace `fetchAndUnblind*Generator` by `txsFetchGenerator` & `utxosFetchGenerator` from latest LDK.

it closes #382 
~~it closes 375~~
it closes #381 

**divide the work of updaters**

batch server update reduces the global number of requests but increases the max number of concurrent requests. On big wallets it leads to several chrome INSUFFICIENT_RESOURCES errors (and sometimes crash the extension). cc @Janaka-Steph 

the idea is to **limit the number of addresses that can be updated at the same time**, this is handled by two hardcoded constants: `MAX_ADDRESSES_TX_GENERATOR = 15` & `MAX_ADDRESSES_UTXO_GENERATOR = 50`. Several generators are then created in the case of large wallet. Then they are launched one after the other.

@tiero please review


package.json
```diff
@@ -36,7 +36,8 @@
     "decimal.js": "^10.2.1",
     "formik": "^2.2.9",
     "google-protobuf": "^3.15.8",
-    "ldk": "^0.5.4",
+    "ldk": "^0.5.7",
+    "liquidjs-lib": "git+https://github.com/louisinger/liquidjs-lib.git#psbt-taproot-support",
```
tiero, line 40: no need for defining this, since we have it in resolution no?

tiero (changes requested):
Shall we add a new input field in Settings > Explorer > Custom for the batch server?

louisinger:
> Shall we add a new input field in Settings > Explorer > Custom for the batch server?

The Custom form was getting a bit big, so I created a new route `/settings/explorer/custom` (with the batch server as well)

src/presentation/settings/explorer.tsx
```diff
@@ -86,26 +93,25 @@ const SettingsExplorer: React.FC = () => {
...
-        <SettingsCustomExplorerForm
-          onDone={() => setCustom(false)}
-          dispatch={dispatch}
-          network={network}
-        />
-      )}
+      <div className="flex justify-end mt-1">
+        <div>
+          <Button type="button" onClick={() => pushToCustomForm()}>
+            Use Custom
```
tiero, line 109: This should be update

tiero: https://www.loom.com/share/00dc8928f3274697980aed90e4aea283?from_recorder=1&focus_title=1

tiero:
I think update button should be there only for custom exploer no? if I switch to Blockstream, it change already and does not give any feedback really

src/application/redux/sagas/updater.ts
```diff
@@ -50,6 +51,9 @@ import { defaultPrecision } from '../../utils/constants';
 import { updateTaxiAssets } from '../actions/taxi';
 import { periodicTaxiUpdater, periodicUpdater } from '../../../background/alarms';
 
+const MAX_ADDRESSES_TX_GENERATOR = 50;
+const MAX_ADDRESSES_UTXO_GENERATOR = 100;
```
Janaka-Steph, line 55: I would use like in the description.
const MAX_ADDRESSES_TX_GENERATOR = 15;
const MAX_ADDRESSES_UTXO_GENERATOR = 50;

On Tdex app 50 txs in one go is still too high

louisinger: I've increased the max addresses in generators due to latest change on batch server (where we batch also the /hex endpoint), @Janaka-Steph did u try to bump ldk using that change?

PS: `MAX_ADDRESSES_*` != number of txs/utxos -> in fact we don't know how many transactions and/or unspents are associated to an address so the performance also depends on wallet state.

tiero: @Janaka-Steph

Janaka-Steph: Yes I should have use v0.5.7

tiero: lowering that constants gives you better results? @Janaka-Steph

Janaka-Steph: Yes 15 and 50 seems ok but I still had one request failure only happening in prod. Need more testing.

PR #409, opened 2022-09-19
address-amount form: handle asset precision

### LINKED ISSUE #406: Handle precision of assets
This asset (38fca2d939696061a8f76d4e6b5eecd54e3b4221c846f24a6b279e79952850a5) has precision of 3 but Marina does not limit decimal places to 3.

louisinger:
* refacto the address-amount form (use `Input` component instead of HTLM input)
* add a formik validation rule checking asset precision in amount field
* move the SEND ALL button **above** the amount field (fix weird margin in case of error) -> need ACK on this @tiero 

@tiero please review

it closes #406

### DIFF
--- a/src/application/redux/containers/send-select-asset.container.ts
+++ b/src/application/redux/containers/send-select-asset.container.ts
@@ -10,7 +10,9 @@ const mapStateToProps = (state: RootReducerState): SendSelectAssetProps => {
   const balances = selectBalances(...selectAllAccountsIDsSpendableViaUI(state))(state);
   const getAsset = assetGetterFromIAssets(state.assets);
   return {
-    balanceAssets: Object.keys(balances).filter(asset => balances[asset] > 0).map(getAsset),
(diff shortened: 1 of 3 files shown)


tiero:
amount less than 0 cant be enter

https://www.loom.com/share/6d9dba2d6a734b90b22be8bb2d353ddf

tiero:
UX still bit clunky:

- allow me to enter more decimals than precision allows
- allow me to enter strings


https://www.loom.com/share/5b8adad6e6ff4251b80579fb2bc1c6cb

tiero:
SEND ALL does not work

https://www.loom.com/share/fb6d343ae63e48389a57470fd91d1792

tiero:
I don't think replacing dots,commas is safe: like if someone is sending thousands, we may want to use the comma as thousands separator.

It's fine to let user type wrong stuff, we simply validate that, the error message should be more useful

PR #415, opened 2022-09-23
payment-error page: RETRY button in case of INVALID_PASSWORD_ERROR

### LINKED ISSUE #414: Send: if password is invalid, the button retry should be present
image

louisinger:
no condition on retry button.

it closes #414 

@tiero please review


louisinger:
UPDATE: end-of-flow now handle the INVALID_PASSWORD error (and let to re-use the Unlock btn if password is invalid)

tiero:
Maybe best to be consistent with the copy of the page?

This is popup

image


image

PR #433, opened 2022-11-17
Replace "poll" updater by websocket subscriber

### LINKED ISSUE #416: Refacto the transactions updater architecture
### How it works now

The current updater is running in background via redux-saga. An update is launch each time an UPDATE_TASK action is dispatched in the store. 
1. [getTxsGenerator](https://github.com/vulpemventures/marina/blob/d160298eb9ed57ab0851974a8df9b090e5a139c9/src/application/redux/sagas/updater.ts#L123-L145) gets the account to update and extract all the addresses (from the current state of the store). Then it launches several LDK generators (txsFetchGenerator) to **fetch and unblind** transactions from the explorer.
2. The results (= the unblinded txs as `TxInterface`) of all generators are [sent to redux-saga channel](https://github.com/vulpemventures/marina/blob/d160298eb9ed57ab0851974a8df9b090e5a139c9/src/application/redux/sagas/updater.ts#L226).
3. That channel is connected to a [txsTransformerWorker](https://github.com/vulpemventures/marina/blob/d160298eb9ed57ab0851974a8df9b090e5a139c9/src/application/redux/sagas/updater.ts#L306-L314) which is a function converting `ldk.TxInterface` (from the channel) to an internal interface named `TxDisplayInterface`.
4. `TxDisplayInterface` has a `transfers` member used in UI to compute the user-friendly informations about the transaction (Inbound & outbound values). In order to do this, the worker is recomputing **all the wallet scripts from state** and send them with the tx in [getTransfers](https://github.com/vulpemventures/marina/blob/d160298eb9ed57ab0851974a8df9b090e5a139c9/src/application/utils/transaction.ts#L395) function.
5. the `TxDisplayInterface` object is the one persisted in redux state. once fetched, unblinded and computed, `TxInterface` is not accessible anymore by the UI.

### The problem

The steps (1) and (4) are really sensitive because they need the wallet scripts. But those scripts come from the accounts `restorerOpts` (the last internal/external indexes) from state. If for any reason, the state is not up to date in background during the execution of step (1) or (4), the `TxDisplayInterface` will be wrong and UI will display bad informations to the user.

### Proposals

I see two refactors to fix the update txs process:

**A:** We persist `TxInterface` in the redux state instead of `TxDisplayInterface`, getTransfers is executed only while needed in the UI. It means the UI reflects exactly the current state value (inbound/outbound are computed when user is looking at tx history, not during the update step).

**B:** Add an event based unblinder worker. Fetching & unblinding would be triggered by the UPDATE_TASK exactly like it works currently. But INCREMENT_*_INDEX actions could also trigger an unblind process, iterating through persisted `TxInterface` in the state and trying to unblind outputs associated with the address' script. 

cc @tiero @altafan

louisinger:
**This PR refacto the way background script is updating (fetching and unblinding) the wallet state (utxos and transactions).**

* no more **UPDATE_TASK** -> all this logic has been drop in `updater.ts`.
* no more updater & unblinding logic in redux-saga -> only "pure" JS function in background page.
* baground page is listening for history change via websocket electrum API.
* `blockchain.scripthash.subscribe` subscriber is set each time we restore or generate an address.
* when a script receive an update `blockchain.transaction.get` & `blockchain.block.header` requests are launch to get the new transaction and handle unspents state management. 
* each update triggers a set of actions to the store (ADD_TX, ADD_UTXO, DELETE_UTXO and CONFIRM_TX actions).
 
please @tiero @altafan @bordalix  review

it closes #416


tiero:
Why you cant use Electrum WS also to fetch the script? So ee remove any dependency from http REST

tiero:
Is the plan to use sockets also for restore in this PR or we move to another DEDICATED PR?

louisinger:
> Is the plan to use sockets also for restore in this PR or we move to another DEDICATED PR? 

Yeah this PR is handling "restore" = subscribe the restored scripts. So you should see the coins/txs of restored mnemonic

louisinger:
> Why you cant use Electrum WS also to fetch the script? So ee remove any dependency from http REST @louisinger

Yeah I've pushed a new commit (and updated PR description). Now the websocket manager is handling fetch & unblind via WS endpoints.

tiero:
If you spend from custom script account, will fail to track it

tiero:
Restore seems to not work, same mnemonic

Marina from stores
image


This branch
image

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,244 @@
...
+    },
+    vin: [],
+    vout: [],
+  };
+
+  for (const input of transaction.ins) {
+    const prevoutTxID = Buffer.from(input.hash).reverse().toString('hex');
+    const vout = input.index;
+
+    if (!input.isPegin) {
```
tiero, line 204: avoid wrapping in curlybraces?

```suggestion
    if (input.isPegin) return;
```

src/domain/ws/observable.ts
```diff
@@ -0,0 +1,39 @@
+type Func = (...args: any[]) => any;
```
tiero, line 1: this looks like coming from here.

https://github.com/nimiq/electrum-client/blob/master/electrum-ws/ElectrumWS.ts

There is a way we can use that library directly? and/or fork to change Liquid responses/data structure?


We should make attribution of code if not

louisinger: Yeah it's a copy paste of the web socket client. We cannot use it due to weird dependency with bitcoinjs. I think easier to just reuse that TS client only.

src/application/redux/reducers/utxos-unspents.ts
```diff
@@ -0,0 +1,277 @@
...
+import type { AccountID } from 'marina-provider';
+import { MainAccountID } from '../../../domain/account';
+import type { TxDisplayInterface } from '../../../domain/transaction';
+import { TxStatusEnum } from '../../../domain/transaction';
+import type { UtxosTransactionsState } from '../../../domain/wallet';
+import { toStringOutpoint } from '../../utils/utxos';
+import * as ACTION_TYPES from '../actions/action-types';
+
+// minimum time utxos are locked (5 minutes)
+const lockedUtxoMinimunTime = 300_000;
```
bordalix, line 12:
```suggestion
const lockedUtxoMinimumTime = 300_000;
```
And do a find & replace project wide

bordalix (approved):

bordalix:
> Restore seems to not work, same mnemonic
> 
> Marina from stores image
> 
> This branch image

Confirm, can replicate.

NACK

tiero:
just nit on folder structure: the websocket client code does not looks like to me a "domain" kind of thing, ie. its not a user domain problem we are solving for him. I would move it to lib/application/utils etc.. (what usually is infrastructure or ports in DDD) cc/ @altafan

altafan:
>just nit on folder structure: the websocket client code does not looks like to me a "domain" kind of thing, ie. its not a user domain problem we are solving for him. I would move it to lib/application/utils etc.. (what usually is infrastructure or ports in DDD)

as we discussed this offline, the idea is that the domain layer contains ALL and ONLY type structures that have be persisted in the repositories. Also, the only secure way to change the internal state of these entities is by calling the methods they expose. Last but not least, every domain should be properly and thoroughly unit-tested to make sure their behavior is the desired one.

bordalix:
@louisinger after [5dc1e81](https://github.com/vulpemventures/marina/pull/433/commits/5dc1e8135f279eb36c18a27382e65f5ad01b6cbd):
- bug 1= when opening the wallet for the first time, it doesn't recover the funds
- bug 2 = restoration is really slow (after 15 minutes is still restoring)
- bug 3 = when I do a restoration, funds go up and down:
   - my real balance is
       - Tether = 1393
       - Fuji USD = 142
       - LBTC = 0.219
   - during restoration, I saw:
       - Tether is correct (maybe because I almost don't use it)
       - Fuji USD going up to 180 and then going down to 142
       - LBTC  going up to 0.69 and then going down to 0.219

tiero:
> * bug 2 = restoration is really slow (after 15 minutes is still restoring)

multi-threading is the way: it's time for web workers in the background script 
https://web.dev/workers-basics/

tiero:
also redux as a discreet overhead, since it uses message passing + JSON de/serialization.

WebWorker should help with CPU-expensive tasks (hashing, unblinding etc..), but I/O like storing in the cache should IMHO just use localStorage in the main thread in the background script.

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,361 @@
...
+  const historyResp = (await ws.request(
+    'blockchain.scripthash.get_history',
+    scriptHash
+  )) as History;
+  const historyDiff = selectHistoryDiff(historyResp)(store.getState());
+  const accountID = selectAccountIDByScriptHash(scriptHash, network)(store.getState());
+  const blindKeyGetter = getPrivateBlindKeyGetter(store, network);
+  const walletScripts = await getAllWalletScripts(store, network);
+
+  for (const { txID, height } of historyDiff.confirmedTxs) {
```
tiero, line 207: this is quite expensive in terms of CPU and blocking. consider moving it in a web worker. It's not vital information to have it available right away, can be updated in background

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,361 @@
...
+    if (height <= 0) continue;
+    const blockHeaderHex = (await ws.request('blockchain.block.header', height)) as string;
+    const blockHeader = deserializeBlockHeader(blockHeaderHex);
+    const action = confirmTx(txID, blockHeader.timestamp, network);
+    store.dispatch(action);
+  }
+
+  // for each new tx which is not in the state, we fetch the transaction and unblind it
```
tiero, line 215: we have plenty of operations here that require redux store getting data, which also may be I/O intensive. If we move to chrome.storage we can decrease the overhead maybe?

Like `blindKeyGetter`, can simply be replaced by passing the full master blinding key and derive it on the go or  if we cache script with the linked blidning key already dervie in storage, we can read sync from there?.

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,361 @@
...
+  for (const { txID, height } of historyDiff.newTxs) {
+    if (selectIsKnownTx(txID)(store.getState())) continue;
+    try {
+      store.dispatch(pushUpdaterLoader());
+      const txFromElectrum = await fetchTx(ws, txID, height > 0 ? height : undefined);
+      const { unblindedTx } = await unblindTransaction(txFromElectrum, blindKeyGetter);
+      const displayTx = toDisplayTransaction(unblindedTx, walletScripts, networks[network]);
```
tiero, line 222: shall this be done in the presentation layer when user opend the tx detail page?

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,361 @@
...
+      }
+
+      const state = store.getState();
+      const allTransactions = selectTransactions(...selectAllAccountsIDs(state))(state);
+
+      // check if we have any new utxos
+      const walletOutputs = unblindedTx.vout
+        .filter((o) => {
+          if (isConfidentialOutput(o)) {
+            return isUnblindedOutput(o);
```
tiero, line 240: why we return always true here just if is unblinded?

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,361 @@
...
+          }
+
+          return walletScripts.includes(o.prevout.script.toString('hex'));
+        })
+        .filter(
```
tiero, line 245: we can avodin two filter which are also bit expensive, if we have a data strcuture in storage that simplify the script existance lookup?

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,361 @@
...
+    const output = transaction.outs[outIndex];
+    txInterface.vout.push({
+      txid: txID,
+      vout: outIndex,
+      prevout: output,
+    });
+
+    // if fee output, add the fee value
+    if (output.script.length === 0) {
+      txInterface.fee = confidential.confidentialValueToSatoshi(output.value);
```
tiero, line 331: You cna use new ElementValue?

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,361 @@
...
+  const txHex = (await ws.request('blockchain.transaction.get', txID)) as string;
+  const transaction = Transaction.fromHex(txHex);
+
+  let blockHeader: BlockHeader | undefined;
+  if (blockheight) {
+    const blockHeaderHex = (await ws.request('blockchain.block.header', blockheight)) as string;
+    blockHeader = deserializeBlockHeader(blockHeaderHex);
+  }
+
+  const txInterface: TxInterface = {
```
tiero, line 287: This TxInterface building is quite expensive too

src/application/redux/selectors/wallet.selector.ts
```diff
@@ -172,3 +200,56 @@ export function selectRestorerOpts<T extends StateRestorerOpts>(account: Account
...
+    newTxs: { txID: string; height: number }[];
+    confirmedTxs: { txID: string; height: number }[];
+  } => {
+    const newTxs: { txID: string; height: number }[] = [];
+    const confirmedTxs: { txID: string; height: number }[] = [];
+
+    for (const { tx_hash, height } of history) {
+      const txInState = selectTransactionByID(tx_hash)(state);
+      // if the tx is in the state, add it in the confirmedTxs array if it's confirmed according to the history
+      if (txInState && !txInState.blockTimeMs && height > 0) {
```
tiero, line 221: just blockHeight is enouhg: timestamp can be lazily loaded for performance

src/application/redux/selectors/wallet.selector.ts
```diff
@@ -172,3 +200,56 @@ export function selectRestorerOpts<T extends StateRestorerOpts>(account: Account
...
+    state: RootReducerState
+  ): {
+    newTxs: { txID: string; height: number }[];
+    confirmedTxs: { txID: string; height: number }[];
+  } => {
+    const newTxs: { txID: string; height: number }[] = [];
+    const confirmedTxs: { txID: string; height: number }[] = [];
+
+    for (const { tx_hash, height } of history) {
+      const txInState = selectTransactionByID(tx_hash)(state);
```
tiero, line 219: if you just keep in chrome.storage the outpoint, any lookup it's much more easier

src/background/deep-restorer.ts
```diff
@@ -0,0 +1,71 @@
...
+      await new Promise((resolve) => setTimeout(resolve, 100));
+    }
+    // run the task
+    this.runningTask = true;
+    try {
+      await restoreAccount(this.store, accountID, network);
+    } finally {
+      this.runningTask = false;
+    }
+  }
```
tiero, line 70: also spawning multiple setTimeouts may increase overhead. Also these tasks would be best to spawn as a service workers

src/background/deep-restorer.ts
```diff
@@ -0,0 +1,71 @@
...
+    const stateRestorerOpts = getStateRestorerOptsFromAddresses(addresses);
+    store.dispatch(setRestorerOpts(accountID, stateRestorerOpts, network));
+  } catch (error) {
+    store.dispatch(
+      setDeepRestorerError(new Error(`deep restore error: ${extractErrorMessage(error)}`))
+    );
+  } finally {
+    store.dispatch(popRestorerLoader());
+  }
+}
```
tiero, line 46: let's try to rewrite this standalone here only using chrome.storage APIs to persist state. (ie. without using redux to dispatch actions and so on)

The CPU intensive operations such as derivation and unblinding, should be spawn in dedicated service worker and then will write the result in chrome.storage

src/background/websocket-manager.ts
```diff
@@ -0,0 +1,361 @@
...
+  )) as History;
+  const historyDiff = selectHistoryDiff(historyResp)(store.getState());
+  const accountID = selectAccountIDByScriptHash(scriptHash, network)(store.getState());
+  const blindKeyGetter = getPrivateBlindKeyGetter(store, network);
+  const walletScripts = await getAllWalletScripts(store, network);
+
+  for (const { txID, height } of historyDiff.confirmedTxs) {
+    if (height <= 0) continue;
+    const blockHeaderHex = (await ws.request('blockchain.block.header', height)) as string;
+    const blockHeader = deserializeBlockHeader(blockHeaderHex);
```
tiero, line 210: this seems something we do not want to block the main thread to just get the timestamp. If it's also done in a dedicated worker, it will then enrich the timestamp field only when ready directly in chrome.storage. UI eventually will show that field only when available, ignoring when undefined, since are still able to draw the UI

tiero:
is this obsolete? @louisinger

louisinger:
> is this obsolete? @louisinger 

Yes, closed by #436

Issue #427, opened 2022-10-06
Custom Script Accounts: lower level APIs


tiero:
# `TL;DR`

```ts
marina.createAccount('my-account@v2')
marina.useAccount('my-account@v2')
marina.getPublicKey() // { publicKey, derivationPath }

// web app generate his own script creation with it
// like using Ionio etc..

marina.importContract({ script, template, constructorParams, derivationPath })
marina.signSchnorr({ sighash, derivationPath })
```

## Reasons

- Adding multiple accounts for little variations of the script template and/or is constructor parameters is a bit of too much effort
- sometimes you do not want to have marina to generate scripts for you, so it's ok to have web apps to import scripts from outside
- This increases possibility mobile apps/hardware wallet could follow this approach

## What happens

- user MUST call `importContract` if want marina to sign/spend it & track coins
- a single account, can have different script templates, one for each public key potentially
- getNextAddress: not clear what should return, **this should be discussed**

Please @louisinger @altafan @bordalix tell me your thoughts

louisinger:
> * Adding multiple accounts for little variations of the script template and/or is constructor parameters is a bit of too much effort
> * sometimes you do not want to have marina to generate scripts for you, so it's ok to have web apps to import scripts from outside

An alternative could be to drop `importContract` and pass the contract as parameter in `getNextAddress` calls (exactly like we did for the `constructorParams` array in fact). Thus, custom script accounts becomes only a BIP32 keys tree derived from the account name: 

```js
marina.createAccount('my-account@v2')
marina.useAccount('my-account@v2')
// I don't need any template/contract to do this:
marina.getPublicKey() // { publicKey, derivationPath }
marina.signSchnorr({ sighash, derivationPath })
```

Then, either the webapp is responsible to generate the script and "send" to marina the generated script. Maybe associated to a derivation path for signing "simple scripts" :thinking: 
```js
marina.registerScript('my-account@v2', scriptDerivationPath, script)
```
Or the webapp use template-based derivation for the next account path:
```js
marina.useAccount('my-account@v2')
marina.getNextAddress({ template, constructorParams }) 
// marina generates the script and map it with next derivation path
```

TL;DR template becomes an address parameter instead of account parameter.
What do you think? @tiero @bordalix

tiero:
[...]

PR #436, opened 2023-01-14
Remove `redux` and `ldk`, refactor using `liquidjs-lib` and electrum websocket endpoint

louisinger:
**This PR removes `redux`, `redux-saga`,  `webext-redux` and `ldk` packages from Marina.**

* replace redux based storage by repositories interfaces in `src/infrastructure`. The repositories classes are using `Browser.storage` to store and access data from local browser extension storage. Repositories can be used everywhere (including React components).
* `src/infrastructure/storage` provides react hooks based on `Browser.storage.onChanged` event listener. It lets to reproduce the behavior of redux `useSelector` hook.
* most of `src/content` and  `src/inject` are unchanged. They are just using Repository interfaces instead of redux state.
* `Account` is now a class able to generate scripts, create confidential addresses and sync the wallet (its a kind of "watch only wallet").
* `SignerService` is a class letting to sign psetv2 using marina Repository interfaces.
* `BlinderService` is a class letting to blind psetv2 using marina Repository interfaces.
* `background-script.ts` is reponsible to update the wallet state using several async services (`Subscriber`, `Updater` and `TaxiUpdater`) listening to `Browser.storage.onChanged` events

_TODO_:
* [x] Fix `wallet/send` flow (psetv2 coin selection bug at choose-fee step).
* [x] Taxi topup support in send flow.
* [x] Custom script `Account` -> related to #427 
* [x] Implement `connect/send` popup.

@bordalix @tiero please review


Janaka-Steph:
I am looking at using React-Query (https://tanstack.com/query/latest) to manage Esplora data, but afaik it returns either `data` or `error`. If I use `ws.batchRequest` to get transactions and one of them fails the whole operation fails. Would it be possible to return successful data, if any, along with promise rejection error inside ElectrumWS onMessage ?
That way I could still display successful requests.

public/manifest.json
```diff
@@ -49,4 +49,4 @@
     },
     "default_title": "Marina"
   }
-}
+}
```
tiero, line 52: Shouldnt be v3? @louisinger

src/domain/account.ts
```diff
@@ -1,264 +1,342 @@
...
+    if (this.walletRepository && updateCache) {
+      // persist the script details in the wallet repository
+      await this.walletRepository.updateScriptDetails(
+        Object.fromEntries(
+          scripts.map((script) => [
+            script.toString('hex'),
+            {
+              accountName: this.name,
+              network: this.network.name as NetworkString,
+              derivationPath: `m/${chain}/${start}`,
```
Janaka-Steph, line 134: Why this derivation path is missing the "coin" segment 84' or 1776'?
Why "chain" is missing the ' ?
Why the index is always `start`?

src/domain/explorer.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const VULPEM_ELECTRS_BATCH_TESTNET = 'https://electrs-batch-testnet.vulpem.com';
+
+export const BlockstreamExplorerURLs: ExplorerURLs = {
+  type: 'Blockstream',
+  webExplorerURL: 'https://blockstream.info/liquid',
+  explorerURL: 'https://blockstream.info/liquid/api',
+  batchServerURL: 'https://electrs-batch-blockstream.vulpem.com',
+  websocketExplorerURL: 'wss://blockstream.info/liquid/electrum-websocket/api',
```
Janaka-Steph, line 19: This comment says the url shouldn't have `/api` https://github.com/Blockstream/esplora/issues/428#issuecomment-1357090919

src/domain/signer.ts
```diff
@@ -0,0 +1,78 @@
...
+    const passwordHash = await walletRepository.getPasswordHash();
+    if (hashPassword(password) !== passwordHash) throw new Error('Invalid password');
+    const encrypted = await walletRepository.getEncryptedMnemonic();
+    if (!encrypted) throw new Error('No mnemonic found in wallet');
+    const decryptedMnemonic = decrypt(encrypted, password);
+    const masterNode = bip32.fromSeed(await mnemonicToSeed(decryptedMnemonic));
+    return new SignerService(walletRepository, masterNode);
+  }
+
+  async signPset(psetBase64: string): Promise<string> {
```
Janaka-Steph, line 28: Maybe better to take a `Pset` directly instead of `psetBase64`?

src/domain/transaction.ts
```diff
@@ -1,57 +1,30 @@
...
-export type UtxosAndTxsByNetwork = Record<NetworkString, UtxosAndTxs>;
-
-export interface UtxosAndTxs {
-  // outpoint string -> UnblindedOutput
-  utxosMap: Record<string, UnblindedOutput>;
-  transactions: TxsHistory;
-}
-
-export type TxsHistory = Record<TxDisplayInterface['txId'], TxDisplayInterface>;
+export type UnblindingData = {
```
Janaka-Steph, line 1: You could use `UnblindOutputResult` type from LiquidJS

Janaka-Steph:
I noticed that we have `txID`, `txid`, `txHash` from LiquidJS. Would be nice to harmonize that.

src/infrastructure/storage/wallet-repository.ts
```diff
@@ -0,0 +1,467 @@
...
+  }
+
+  private async getAllAccountsNames(): Promise<Array<string>> {
+    const wholeStorage = await Browser.storage.local.get(null);
+    return Object.entries(wholeStorage)
+      .filter(([key, value]) => AccountKey.is(key) && value !== null)
+      .map(([key]) => AccountKey.decode(key)[0]);
+  }
+
+  static LOCKTIME = 5 * 60 * 1000; // 5 minutes
```
Janaka-Steph, line 423: Why do you lock outputs for 5 minutes?

Janaka-Steph: Instead of using an `until` timestamp, I just do this at coin selection:
```
const lockedOutpoints = selectedUtxos.map((utxo) => get().lockOutpoint(utxo));
await sleep(60_000);
lockedOutpoints.forEach((outpoints) => get().unlockOutpoint(outpoints));
```
Am I missing something?

src/extension/wallet/send/choose-fee.tsx
```diff
@@ -0,0 +1,342 @@
...
+        ]);
+      }
+
+      const feeRate = 0.001;
+
+      const feeAmount = Math.floor(updater.pset.unsignedTx().virtualSize() * feeRate);
+
+      if (recipient.asset === networks[network].assetHash && updater.pset.outputs.length > 1) {
+        // subtract fee from change output
+        updater.pset.outputs[1].value = updater.pset.outputs[1].value - feeAmount;
```
Janaka-Steph, line 180: Does it mean that we can substract fee from change like in Bitcoin OR adding a specific fee output?
Is it new to Pset v2 ?

louisinger: No we must have a fee output in every cases. this line update the change amount if it is an L-BTC one. It prevents reselecting utxos if we have a change covering the fees.

Janaka-Steph: In current code fee output is only added in the reselect `else` block

src/domain/signer.ts
tiero: Internally we can avoid roundtrip de/serialization

public/manifest.json
louisinger: as the new updater is a pretty huge refacto, I thought it would be better to validate and discuss it first and then migrate to v3 (in a another PR maybe?). 

PS: theorically, it's totally possible to write a v3 manifest and tell webpack to build both v3 & v2 (I'm thinking of firefox)

src/domain/account.ts
louisinger: > Why this derivation path is missing the "coin" segment 84' or 1776'?

The "coin" section is a part of the _hardened_ path and `Account` class does not have access to any private keys (so unable to derive hardened indexes). The coin section and all others hardened ones are derived at the account creation step and pass to `Account` class via the account masterPublicKey (xpub).

> Why "chain" is missing the ' ? 

As far as I know, chain should not be hardened.

> Why the index is always `start`?

nice catch, fixed, now is `start + index`.

src/domain/signer.ts
louisinger: this idea is also to use this `SignerService` in provider `signPset` method (where pset is passed as base64)

src/domain/transaction.ts
louisinger: right. I made this on purpose, I wanted to have a JSON-friendly type in order to drop our custom (de)serializers made for the `Buffer` type.

src/infrastructure/storage/wallet-repository.ts
louisinger: The main advantage to use a cached `until` timestamp is that if the process stops for any reason, we know that our outpoint will be unlocked anyway. This is perfect for the Marina context cause we don't know if the user will stop suddenly the extension/browser/popup.

src/domain/signer.ts
tiero: You can serialize/deserialize directly in the handler, leaving Pset data structure internally no?

src/extension/wallet/send/choose-fee.tsx
```diff
@@ -172,9 +176,13 @@ const ChooseFee: React.FC = () => {
...
 
-      const feeRate = 0.001;
+      const chainSource = await appRepository.getChainSource(network);
+      if (!chainSource) {
+        throw new Error('chain source not found, cannot estimate fee');
+      }
+
 
-      const feeAmount = Math.floor(updater.pset.unsignedTx().virtualSize() * feeRate);
+      const feeAmount = 360;
```
Janaka-Steph, line 185: So the fee amount is just fixed now?

src/extension/components/button-transaction.tsx
```diff
@@ -0,0 +1,188 @@
...
+import Button from './button';
+import Browser from 'webextension-polyfill';
+import {
+  appRepository,
+  useSelectNetwork,
+  walletRepository,
+} from '../../infrastructure/storage/common';
+import { makeURLwithBlinders } from '../../utils';
+
+function txTypeFromTransfer(transfer?: number): TxType {
```
Janaka-Steph, line 22: Renaming suggestion: `txTypeFromTransferAmount(transferAmount?: number): TxType`

src/extension/components/button-transaction.tsx
```diff
@@ -0,0 +1,188 @@
...
+        if (!data || !data.blindingData) continue;
+        if (data.blindingData.asset === assetSelected.assetHash) {
+          transferAmount += data.blindingData.value;
+        }
+      }
+
+      for (let inIndex = 0; inIndex < transaction.ins.length; inIndex++) {
+        const input = transaction.ins[inIndex];
+        const data = await walletRepository.getOutputBlindingData(
+          Buffer.from(input.hash).reverse().toString('hex'),
```
Janaka-Steph, line 70: `input.hash` is already a Buffer. Any reason to use Buffer.from() ?

louisinger: it prevents mutation on `input.hash` (reverse does not copy the buffer)

Janaka-Steph: Ok thanks

src/domain/explorer.ts
```diff
@@ -0,0 +1,52 @@
...
+  wsURL: 'ws://localhost:3001',
+  websocketExplorerURL: 'ws://localhost:1234',
+};
+
+export const MempoolExplorerURLs: ExplorerURLs = {
+  type: 'Mempool',
+  webExplorerURL: 'https://liquid.network',
+  explorerURL: 'https://liquid.network/api',
+  batchServerURL: 'https://electrs-batch-mempool.vulpem.com',
+  websocketExplorerURL: 'wss://esplora.blockstream.com/liquid/electrum-websocket/api',
```
Janaka-Steph, line 43: Mempool doesn't have endoint?

src/domain/account.ts
```diff
@@ -1,264 +1,318 @@
...
+  private chainSource: ChainSource;
+  private network: networks.Network;
+  private node: BIP32Interface;
+  private blindingKeyNode: Slip77Interface;
+  private walletRepository: WalletRepository;
+  readonly name: string;
+
+  static BASE_DERIVATION_PATH = "m/84'/1776'/0'";
+  static BASE_DERIVATION_PATH_LEGACY = "m/84'/0'/0'";
+  static BASE_DERIVATION_PATH_TESTNET = "m/84'/1'/0'";
```
Janaka-Steph, line 81: In LDK (so current apps), the default base derivation path is the same for testnet and mainnet. Not anymore.
What is the reasoning to change that?

tiero: To be compatible with Green

src/domain/account.ts
```diff
@@ -1,264 +1,318 @@
...
+
+    const lastUsed = await this.getLastUsedIndexes();
+
+    const walletChains = [0, 1];
+    for (const i of walletChains) {
+      const isInternal = i === 1;
+      let batchCount = isInternal ? lastUsed.internal : lastUsed.external;
+      let unusedScriptCounter = 0;
+
+      while (unusedScriptCounter <= gapLimit) {
```
Janaka-Steph, line 205: I think there is a bug here. We loop one extra time when having no histories.
Say lastUsedExternalIndex is 3 and gapLimit is 3, we derive scripts/histories for 3 to 6, then  unusedScriptCounter is 3, unusedScriptCounter <= gapLimit is true, we derive again for 6 to 9.

Should it be `unusedScriptCounter < gapLimit` instead?
@louisinger

src/domain/explorer.ts
tiero: No, they wont expose it without an enterprise plan anyway

src/utils.ts
```diff
@@ -0,0 +1,410 @@
...
+      return 1;
+  }
+  return 0;
+}
+
+const INPUT_BASE_SIZE = 40; // 32 bytes for outpoint, 4 bytes for sequence, 4 for index
+
+function txBaseSize(inScriptSigsSize: number[], outNonWitnessesSize: number[]): number {
+  const inSize = inScriptSigsSize.reduce((a, b) => a + b + INPUT_BASE_SIZE, 0);
+  const outSize = outNonWitnessesSize.reduce((a, b) => a + b, 0) + 33 + 9 + 1 + 1; // add unconf fee output size
```
Janaka-Steph, line 124: Can use the constant `outNonWitnessesSize.reduce((a, b) => a + b, 0) + FEE_OUTPUT_SIZE`

src/domain/account.ts
louisinger: Both of them produce the same restore result on my wallet. I would keep the "=" (better to test 1 additional batch instead of missing some txs. Note that if the next tested batch does not have histories, the scripts won't be restored).

src/utils.ts
louisinger: This should not be there, there is a new `withFeeOutput` argument handling that case in `estimateVirtualSize` function. fixed :heavy_check_mark:

louisinger:
**UPDATE**
Latest commits add full support of Ionio accounts. Some small [breaking changes in marina-provider](https://github.com/vulpemventures/marina-provider/pull/30) has been introduced. The [docs.vulpem.com update](https://github.com/vulpemventures/docs.vulpem.com/pull/18) describes them and adds an example related to Ionio account.

It should be enough to test and discuss new APIs @tiero @bordalix 

I keep this PR as draft, some required WIP:
- [ ]  Add tests of the application layer (`Account`, `SignerService` and `BlinderService`).
- [ ] `SignerService` & `BlinderService` should be able to sign/blind psetv0.
- [ ] Let the user download a restoration file with Ionio account data in `settings/accounts`.
- [ ] Let the user upload restoration files in `onboarding/restore`.

src/utils.ts
```diff
@@ -118,10 +118,11 @@ function estimateScriptSigSize(type: ScriptType): number {
 }
 
 const INPUT_BASE_SIZE = 40; // 32 bytes for outpoint, 4 bytes for sequence, 4 for index
+const UNCONFIDENTIAL_OUTPUT_SIZE = 33 + 9 + 1 + 1; // 33 bytes for value, 9 bytes for asset, 1 byte for nonce, 1 byte for script length
```
tiero, line 121: https://github.com/ElementsProject/elements/blob/master/doc/elements-tx-format.md

louisinger:
[marina-0.4.14.zip](https://github.com/vulpemventures/marina/files/10768468/marina-0.4.14.zip)

The latest version adds Ionio accounts restoration (download a file in settings/accounts and upload it in onboarding/wallet-restore form) + Liquid Taxi support.

tiero:
I would put in the placeholder an example of what we need. ie. `https://blockstream.info` and `wss://vulpem.com`

tiero:
I would add also the base derivation path for each account listed here

image

tiero:
If possible I would put the `Updating...` in the navbar during the restoration/fetching of history

tiero:
Reason why these assets have no balance? not even 0

image


Also can we scale the font so we can at least fit the text? 
image

tiero:
Best to write Unconfirmed here, maybe as tag using the red or yellow background 
image

tiero:
The sending flow has some glitches:

- for some reason was giving to a perfect valid address as not valid. To make it work you need to copy paste second time.
- I dont understand why is taking do much time when selecting l-btc as asset to pay for fees: is needed all of this to do what exactly? coin-selection?

https://www.loom.com/share/ab92c4fd7a8e44e6a310f131df51df31

src/content/marina/marinaBroker.ts
```diff
@@ -380,28 +397,42 @@ export default class MarinaBroker extends Broker<keyof Marina> {
...
-            results.push({
-              ...coin,
-              txid: coin.txID,
-              scriptDetails,
-              witnessUtxo,
-            });
+            // if need unblinding, unblind, cache result, and add to the list
+            const unblinded = (await unblinder.unblind(utxo.witnessUtxo)).at(0);
+            if (unblinded instanceof Error || !unblinded) {
+              console.warn('unblinding failed', unblinded);
```
tiero, line 424: Shall we schedule a retry instead of only logging?

tiero: or maybe in this case throwing an error back to the requester

louisinger: I made it on purpose because the `Utxo` type also have a `blindingPrivateKey` field & `blindingData` are optional thus it's also possible to generate them or use the blindingPrivateKey to unblind the input while spending the utxo

tiero: So ok you saying its fine to return the wrong data (ie missing the blindingData), as the user can retry to get more recent coins?

tiero:
nit: overlay background does not cover all extension

image

src/content/marina/marinaBroker.ts
louisinger: I mean most of the time getCoins() is used with coin select (cause we already have getBalances, getTxs for other types of data) so I think it's ok to return the utxo even if marina was not able to unblind it using masterBlindingKey (we don't know, maybe the app is using a special blinding process)

bordalix:
L-BTC balance (on the assets list) should be '0' and not absent as is now:

Screenshot 2023-03-14 at 14 56 43

bordalix:
Date area needs to have less width or else value of the tx is wrapped:

Screenshot 2023-03-14 at 15 01 15

bordalix:
I'm getting blinding errors on mempool and blockstream when seeing a contract creation tx:

https://liquid.network/testnet/tx/0e82f8fca8f968bff3b78c64d7c0f6bb18f94f3d331861c486a770c300143746#blinded=84897,144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49,d6fd0b0dda248b46903ed123a6c0bf7dbb55082f671033289bd3a472434a9bc7,fd1f57f5dfe08560e74c05abd97e3b5635c5831867ace8ed65b920cf2d3cbcf9,21680168,144c654344aa716d6f3abcc1ca90e5641e4e2a7f633bc09fe3baf64585819a49,bfeb9a4d8e1de578bb014e264a21c93715cc6dc4233bc3c948dc2cfcd881e48b,1c923c8f54bd7237274b1a0ae9aead67affb0c4f33c748394e41f8725765ba5c,1100000000,0d86b2f6a8c3b02a8c7c8836b83a081e68b7e2b4bcdfc58981fc5486f59f7518,0000000000000000000000000000000000000000000000000000000000000000,0000000000000000000000000000000000000000000000000000000000000000

Screenshot 2023-03-14 at 15 14 02

Screenshot 2023-03-14 at 15 13 52

bordalix:
1. Create contract
2. Export `marina-ionio-restoration.json`
3. In marina, restore account
4. Restore with same mnemonic and json file from 2
5. Contracts has status "unknown" (I will check why)
6. When I try to close the contract, I get this error:

Screenshot 2023-03-14 at 15 28 14

tiero:
Can you please try to "Restore account" from a wallet already initialized? For some reason I always get invalid password that way @louisinger

tiero:
Shall we add a button into either Accounts or in Deep Restore to be able to import the restoration file if needed?

tiero:
image

tiero:
When L-BTC is part of tx and used as fee input to cover the fee output, it should not be shown in the L-BTC history?

image

bordalix:
When I try to send all LBTC to myself I get a `min relay fee not met, 0 < 134` 

https://user-images.githubusercontent.com/9318412/225405661-475dbab3-9029-4ef5-8156-6c9a88f86291.mov

Also, Retry button does nothing.

bordalix:
Log out NEVER works on first click:

https://user-images.githubusercontent.com/9318412/225659918-5084e209-7755-4f6d-be6e-38155e8644cd.mov

PR #446, opened 2023-03-22
correctly load the state in address-amount.tsx form

### LINKED ISSUE #439: `address-amount` form should display the current state of sendFlowRepository
related to #436 

if user has fill the address and/or amount form inputs. It should see their values while re-opening the marina popup.

louisinger:
it closes #439 

`wallet/send/address-amount` shows the persisted state (address and/or amount set up by the user).

@tiero please review


bordalix:
tACK

PR #448, opened 2023-03-22
move `useSelectTransations` logic to `StorageContext`

### LINKED ISSUE #437: Home page asset list should only show owned asset
related to #436 

Currently while creating a fresh new wallet, home page (route: "/") shows the taxi assets (Vulpem EUR & Vulpem USD). 
it should only show asset if we have transactions related with it.

louisinger:
This PR drops the `useSelectTransactions` hook used in `transactions` view and compute the transactions list from repositories inside the `StorageContext` component. It lets to:
* speed up the transactions list rendering
* speed up the "transfer amount" computing
* be consistent with utxos list (also computed in StorageContext)

the PR also use that newly cached transactions list to filter the asset in `Home` view. it closes #437 

@tiero @bordalix please review this

### DIFF
--- a/src/application/unblinder.ts
+++ b/src/application/unblinder.ts
@@ -6,7 +6,7 @@ import type { Output } from 'liquidjs-lib/src/transaction';
 import { SLIP77Factory } from 'slip77';
 import type { AppRepository, AssetRepository, WalletRepository } from '../domain/repository';
 import type { UnblindingData } from '../domain/transaction';
-import { fetchAsset } from './utils';
+import { assetIsUnknown, fetchAssetDetails } from './utils';
 
 const slip77 = SLIP77Factory(ecc);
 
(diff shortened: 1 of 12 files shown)


tiero (commented):
It may be a non-issue, but does having these injected via Context causing re-rendering al all child components everytime a new tx is computed?

louisinger:
> It may be a non-issue, but does having these injected via Context causing re-rendering al all child components everytime a new tx is computed?

From my React understanding, only the components using the props `cache.transactions` will be re-rendered

tiero:
> nly the components using the prop

yep, I did not review the code but want to make sure we confirm we not introducing unnecessary re-renders

louisinger:
> yep, I did not review the code but want to make sure we confirm we not introducing unnecessary re-renders

To be honest if think it's equivalent in terms of react rendering but for sure it reduces the number of repository calls

tiero:
Can you fix conflicts @louisinger ?

PR #450, opened 2023-03-23
Added precision to amount formatting in spend.tsx.

### LINKED ISSUE #449: Registered asset amounts are not being formatted with precision in spend.tsx
When the `marina.sendTransaction` is called from the api, the popup from sign.tsx doesn't format the amount with the precision found within the registry

The asset in the picture below has a precision of 0 but is shown with a precision of 8:
Screenshot 2023-03-23 at 5 53 34 AM

Randy808:
I also included a change in utils.ts's `fetchAsset` to explicitly check whether the precision is null or undefined before assigning the default value of 8 because the valid precision value of 0 always falls into the falsy case and can never be set.

Closes #449

### DIFF
--- a/src/extension/popups/spend.tsx
+++ b/src/extension/popups/spend.tsx
@@ -36,6 +36,11 @@ const ConnectSpend: React.FC = () => {
     return assetInfo ? assetInfo.ticker : asset.slice(0, 4);
   };
 
+  const getPrecision = (asset: string) => {
+    const assetInfo = cache?.assets.find((a) => a.assetHash === asset);
+    return assetInfo ? assetInfo.precision : 8;
+  };
+
   const handleModalUnlockClose = () => showUnlockModal(false);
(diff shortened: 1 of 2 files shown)


src/extension/popups/spend.tsx
```diff
@@ -103,7 +108,7 @@ const ConnectSpend: React.FC = () => {
             {spendParameters.addressRecipients.map((recipient, index) => (
               <div key={index}>
                 <div className="container flex justify-between mt-6">
-                  <span className="text-lg font-medium">{fromSatoshi(recipient.value)}</span>
+                  <span className="text-lg font-medium">{fromSatoshi(recipient.value, getPrecision(recipient.asset))}</span>
```
bordalix, line 111: To remove prettier errors

```suggestion
                  <span className="text-lg font-medium">
                    {(fromSatoshi(recipient.value), getPrecision(recipient.asset))}
                  </span>
```

PR #453, opened 2023-03-30
skip OP_RETURN scripts while computing transfer

### LINKED ISSUE #444: Close contract tx never appear in tx list
1. Initial funding (+234)
2. Borrow FUSD (+7)
3. Close contract (-7) missing
4. Borrow FUSD (+8)
5. Close contract (-8) missing

Screenshot 2023-03-21 at 15 58 05

louisinger:
This PR ensures that we are not adding the burn outputs value to "transfer" amount. 

it closes #444 

### DIFF
--- a/src/domain/transaction.ts
+++ b/src/domain/transaction.ts
@@ -1,5 +1,4 @@
-import { networks, Transaction } from 'liquidjs-lib';
-import { confidentialValueToSatoshi } from 'liquidjs-lib/src/confidential';
+import { AssetHash, ElementsValue, networks, script, Transaction } from 'liquidjs-lib';
 import type { BlockHeader } from './chainsource';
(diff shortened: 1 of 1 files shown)


tiero:
Shouldn't be shown as "burn" output instead of skip?

louisinger:
> Shouldn't be shown as "burn" output instead of skip?

Wdym by "output"? currently a burn tx of 25 L-BTC provided by your wallet will be shown as a -25 L-BTC (withdraw type). Before the PR it was shown as a self-transfer (+0 L-BTC)

tiero:
> EDIT: do u mean add an extra data "burnAmount" = the total number of OP_RETURN sats in the tx ?

Yep I meant that, but it's a new feature indeed, since as you say it was shown as a self-transfer

PR #464, opened 2023-04-10
show zero when there's no balance

### LINKED ISSUE #463: When balance is 0, no value is present
On FUSD the balance should be '0' and not empty:

Screenshot 2023-04-10 at 12 11 34

bordalix:
closes #463 

@tiero please review


bordalix:
> I would make the default return value of the formatFunction to be zero instead

Like this? [2e85b8a](https://github.com/vulpemventures/marina/pull/464/commits/2e85b8a4be7b6fcf48d44d06b037cb30e55c31eb)

PR #467, opened 2023-04-12
Fix manifest v2 extension & release v2 artifact

louisinger:
### This PR fixes v2 version of the extension & adds a release artifact.

- `dist` & `web-ext-artifacts` folders are now divided into `v2` & `v3` directories, depending on which version of the extension you are building.
- add a `web-ext:build:v3` script in pkg.json
- `main.yml` CI is now testing v2 AND v3 build & bundle.
- `release.yml` now upload two artifacts: `marina-v3-{version}.zip` & `marina-v2-{version}.zip`.

_Unfortunately it's not possible to run the playwright tests under Firefox cause playwright supports extension context only on Chrome-based browsers._

please @tiero review this

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 on:
   push:
     tags:
-      - "v*"
+      - 'v*'
 
 name: Release
 
@@ -20,7 +20,7 @@ jobs:
       # setup nodejs
       - uses: actions/setup-node@v2
         with:
-          node-version: "16"
+          node-version: '16'
 
       # get current  tag
       - name: tag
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -43,30 +43,59 @@ jobs:
...
+
+      # Build extension v2
+      - name: Build extension manifest v2
+        run: |
+          yarn build
           yarn web-ext:build
 
       # Release to github
-      - name: Create Release
+      - name: Create Release (manifest v2)
```
tiero, line 82: I think the release must be only one, then you upload multiple zips to the same one

.github/workflows/release.yml
```diff
@@ -72,22 +78,6 @@ jobs:
           asset_name: marina-v3-${{ steps.next_version.outputs.version }}.zip
```
tiero, line 78: here you uplaoding only the v3?

louisinger: Yep, asset_path is: `web-ext-artifacts/v3/......zip`

tiero: we need to upload both v2 and v3 no?

tiero: nvm I see the extra step later

PR #484, opened 2023-06-16
reverse blinding factors while encoded as hex string

louisinger:
This PR ensures that blinders are encoded as **reversed string**. It may be a critical breaking change for web apps using liquidjs-lib `Blinder` (should be released with a proper version management in marina-provider).

### DIFF
--- a/src/application/utils.ts
+++ b/src/application/utils.ts
@@ -1,3 +1,11 @@
 export function h2b(hex: string): Buffer {
   return Buffer.from(hex, 'hex');
 }
+
+export function b2hReversed(buffer: Buffer): string {
+  return Buffer.from(buffer).reverse().toString('hex');
+}
+
+export function h2bReversed(hex: string): Buffer {
+  return h2b(hex).reverse();
+}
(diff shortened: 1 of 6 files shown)


src/application/utils.ts
```diff
@@ -1,3 +1,11 @@
 export function h2b(hex: string): Buffer {
   return Buffer.from(hex, 'hex');
 }
+
+export function b2hReversed(buffer: Buffer): string {
+  return Buffer.from(buffer).reverse().toString('hex');
```
bordalix, line 6: Why
`Buffer.from(buffer).reverse().toString('hex');`
instead of
`buffer.reverse().toString('hex');`
?

bordalix (approved):

src/application/utils.ts
louisinger: `reverse()` function mutates the original buffer/array. It led to lot of errors in the past. `Buffer.from` is a trick to "copy" the buffer and create a new one (thus reverse is not applied on `buffer`)

PR #494, opened 2023-10-18
SignerService: request signature using BIP32Derivation field

louisinger:
TL;DR

* `SignerService` will fallback to `input.BIP32Derivation[0]` field to derivate the signing key if `witnessUtxo.script` is unknwown.
* Add a test case signing a simple pset without witnessUtxo.

_Note that the fields `fingerPrint` and `path` are not used because Marina is mapping derivation path to script. Instead this PR uses a trick to re-compute the P2WPKH script from `publicKey` field_. In other words your only need to set up a marina public public in the input to request a signature.

try it yourself: [marina-0.5.4.zip](https://github.com/vulpemventures/marina/files/13033604/marina-0.5.4.zip)

### DIFF
--- a/src/application/signer.ts
+++ b/src/application/signer.ts
@@ -12,6 +12,7 @@ import {
   bip341,
   networks,
   Updater,
+  payments,
 } from 'liquidjs-lib';
 import { mnemonicToSeed } from 'bip39';
 import type { AppRepository, WalletRepository } from '../domain/repository';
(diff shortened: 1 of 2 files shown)


tiero:
Can you attach a zip to test?

jamesdorfman:
Great work, thank you. tACK 5888ff3ae1a50203274350402aee0edf71d487cd.

Marina successfully signed my PSET when I provided a Bip32Derivation.

PR #513, opened 2023-12-21
Keep state for LN > LQ swaps

### LINKED ISSUE #510: LN > LQ swap must lock the screen state until completed
If user closes popup during a LN to LIQUID swap, once he reopens we  user should complete the swap if initiated before.

There is a timeout for the LN payment, we should take that into consideration before attempting to claim after use reopens popup

bordalix:
This PR:
- splits receiving in lightning into 2 differet routes:
  - insert amount
  - show invoice
- stores on local storage amount inserted and swap data
- if the user already inserted an amount, when coming back will see that page with previous amount
- if user closes the popup before ending the claiming (ie receiving LN), he can now re-open it and the app will remember and finalize the claim tx

closes #510 

@tiero please review


src/extension/wallet/receive/lightning-enter-amount.tsx
```diff
@@ -48,135 +45,95 @@ const LightningAmount: React.FC = () => {
...
+    fetchAmount().catch(console.error);
+  }, []);
+
+  const makeSwap = async (): Promise<SwapData> => {
+    // we will create an ephemeral key pair:
+    // - it will generate a public key to be used with the Boltz swap
+    // - later we will sign the claim transaction with the private key
+    const claimPrivateKey = randomBytes(32);
+    const claimKeyPair = ECPairFactory(ecc).fromPrivateKey(claimPrivateKey);
+    const claimPublicKey = claimKeyPair.publicKey;
```
tiero, line 66: Shall we derive a well-known public key instead? Just thinking what if the user removes/uninstalls marina before completing the swap?

bordalix: I guess we don't loose anything by using it.
Now fixing merge conflicts 😬

bordalix: @tiero in a claim transaction we need the full key pair: public and private key.
If we derive the public key from a marina address, that means we'll need to ask the user for his password in order to have the private key (need to sign the claim transaction).

From a UX perspective, I prefer NOT to ask users for a password for them to RECEIVE.
I don't think is fair for thousands of users the need to insert a password to receive just because one guy could remove/uninstall marina before completing the swap.

What do you think?

tiero: Also, I guess if you do not claim, the LN payment will simply be refunded to the sender if never claimed on-chain passing the right preimage and the righ signature.

So agreed, let's make it random and store it.

PR #515, opened 2023-12-26
boltz: add Refund method

tiero:


src/pkg/boltz.ts
```diff
@@ -281,6 +305,68 @@ export class Boltz implements BoltzInterface {
...
+    const signedPset = this.signPset(blindedPset, refundKeyPair);
+
+    const finalizer = new Finalizer(signedPset);
+
+    finalizer.finalizeInput(0, (inputIndex, pset) => {
+      return {
+        finalScriptSig: undefined,
+        finalScriptWitness: witnessStackToScriptWitness([
+          pset.inputs[inputIndex].partialSigs![0].signature,
+          Buffer.of(), //dummy preimage
```
tiero, line 362: The only difference between a refund and claim tx, is that you can pass an empty preimage

https://github.com/BoltzExchange/boltz-core/blob/master/lib/liquid/swap/Refund.ts#L6

tiero:
Fell free to `git cherry-pick` this [commit](https://github.com/vulpemventures/marina/pull/515/commits/95b5903275aab404cfad9d99398db1a5e3415256) to test it live with UI

src/pkg/boltz.ts
tiero: false: you also need the heightLocktime to be set on the input :)

PR #521, opened 2024-01-03
Remembers invoice

bordalix:
This PR makes Marina remember previous invoices and don't ask for a submarine swap again.

Implements a new updater strategy for refundable tx (tx from failed swaps):
- these swaps can be in 2 different scenarios:
  - they were broadcasted (ie, have a txid)
    - remove them if they were swepts by Boltz (not refundable anymore because were spent)
  - user didn't went through and didn't broadcasted the tx
    - kept in cache cause we can't ask Boltz for another swap with the same invoice, so we will use this values if that happens
    - after invoice expiration date these swaps are removed from storage 

Fixes bugs on tx list:
  - Wrong 0 
  - Refund button was available on tx not confirmed

@tiero please review

### DIFF
--- a/src/application/updater.ts
+++ b/src/application/updater.ts
@@ -88,6 +88,14 @@ export class UpdaterService {
     }
   }
 
+  // check if something changed on submarine swaps made in the past
+  // these swaps can be in 2 different scenarios:
+  // - were broadcasted (have a txid)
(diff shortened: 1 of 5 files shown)


tiero:
tACK re-using invoice like @altafan reported

didn't tested let invoice expiring, but should be safe if it happens

